PR #29, opened 2021-10-03
Added has_role method for member

ajratnam:

- [ ] Breaking code change

## Description

Added the `has_role` method which is marked as `todo` in the Trello Board

## Changes

- Added `has_role` to `member` object in `user.py`

I ran the code and it works.

### DIFF
--- a/dis_snek/models/discord_objects/user.py
+++ b/dis_snek/models/discord_objects/user.py
@@ -328,6 +328,19 @@ class Member(DiscordObject, _SendDMMixin):
             role = role.id
         return await self._client.http.remove_guild_member_role(self._guild_id, self.id, role, reason=reason)
 
+    async def has_role(self, *roles: Union[Snowflake_Type, Role]) -> bool:
+        """
+        Checks if the user has the given role(s)
+        Args:
+            roles: The role(s) to check whether the user has it.
+        """
+        for role in roles:
+            if isinstance(role, Role):
+                role = role.id
+            if role not in self._role_ids:
+                return False
+        return True
+
     async def kick(self, reason: str = MISSING):
         """
         Remove a member from the guild.
--- a/dis_snek/models/discord_objects/user.py
+++ b/dis_snek/models/discord_objects/user.py
@@ -337,9 +337,8 @@ class Member(DiscordObject, _SendDMMixin):
             roles: The role(s) to check whether the user has it.
         """
         for role in roles:
-            if isinstance(role, Role):
-                role = role.id
-            if str(role) not in self._role_ids:
+            role_id = to_snowflake(role)
+            if role_id not in self._role_ids:
                 return False
         return True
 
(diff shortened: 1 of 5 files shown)


ajratnam:
Please let me know if I have to change anything.

dis_snek/models/discord_objects/user.py
```diff
@@ -328,6 +328,19 @@ async def remove_role(self, role: Union[Snowflake_Type, Role], reason: str = MIS
...
+    async def has_role(self, *roles: Union[Snowflake_Type, Role]) -> bool:
+        """
+        Checks if the user has the given role(s)
+        Args:
+            roles: The role(s) to check whether the user has it.
+        """
+        for role in roles:
+            if isinstance(role, Role):
+                role = role.id
+            if str(role) not in self._role_ids:
```
benwoo1110: Why convert to str here? role id should be an integer since it's a snowflake. (unless we didn't convert properly somewhere prior)

benwoo1110 (commented):
Generally looks good, id prefer the typo fixes to be separate commits but it's fine.

dis_snek/models/discord_objects/user.py
ajratnam: I also did think it was int first and ran my checks and it failed, but then i checked the `_role_ids` it was str, so i thought it was supposed a str, and converted it to str.

`print(type((await bot.get_member(894251257378783283,850312993496891392))._role_ids[0]))` -> `str`

benwoo1110: yea forgot to convert, just fixed it in latest commit should be all int now.

ajratnam: @benwoo1110  Yes I think conversion isn't proper, `missing the to_snowflake converter`, I added ` converter=lambda ids:[to_snowflake(id) for id in ids]` (Is there any other list converter? the list_converter didn't work) and its all ints in the __role_ids

benwoo1110: Hmm I get all ints after commit https://github.com/LordOfPolls/dis_snek/commit/06b8995a7e9abf144bbeeddfb3dc7d877110d18b Either you haven't update your branch or something you are doing different.

LordOfPolls:
[`member.has_role` - Check if member has a role(s)](https://trello.com/c/Ji8KPRM6/82-memberhasrole-check-if-member-has-a-roles)

LordOfPolls (commented):
Tested and this works. Once @benwoo1110's review is resolved, this can be merged. Good work 👏

As an aside, in future, please try and keep doc changes (your typo fixes) to their own PR. Mixing PR goals is bad practice

dis_snek/models/discord_objects/user.py
ajratnam: Oh, that commit came after I committed, so i didn't know

ajratnam:
> Tested and this works. Once @benwoo1110's review is resolved, this can be merged. Good work 👏
> 
> As an aside, in future, please try and keep doc changes (your typo fixes) to their own PR. Mixing PR goals is bad practice

Noted.

ajratnam:
I updated the code now and tested it. And it works.

benwoo1110 (approved):
LGTM

PR #31, opened 2021-10-10
Add has_permission() method to member

Kigstn:

- [ ] Breaking code change

## Description

This PR adds the `has_permission()` method to the member class. This includes documentation about it. 
Taken from https://trello.com/c/Olhwmubt/77-helper-methods

## Changes

- add `has_permission()` method to member

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/user.py
+++ b/dis_snek/models/discord_objects/user.py
@@ -286,7 +286,7 @@ class Member(DiscordObject, _SendDMMixin):
         """
 
         # Get the user's permissions
-        guild_permissions = self.guild_permissions()
+        guild_permissions = await self.guild_permissions()
 
         # Check all permissions separately
         for permission in permissions:
--- a/dis_snek/models/discord_objects/user.py
+++ b/dis_snek/models/discord_objects/user.py
@@ -268,18 +268,19 @@ class Member(DiscordObject, _SendDMMixin):
 
     async def has_permission(self, *permissions: Permissions) -> bool:
         """
-        Checks if the member has all the given permission(s)
+        Checks if the member has all the given permission(s).
 
         ??? Hint "Example Usage:"
-            This returns `True`:
+            Two different styles can be used to call this method.
+
             ```python
-            await admin.has_permission(Permissions.ADMINISTRATOR)
+            await member.has_permission(Permissions.KICK_MEMBERS, Permissions.BAN_MEMBERS)
+            ```
+            ```python
+            await member.has_permission(Permissions.KICK_MEMBERS | Permissions.BAN_MEMBERS)
             ```
 
-            This returns `False`:
-            ```python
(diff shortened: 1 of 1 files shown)


dis_snek/models/discord_objects/user.py
```diff
@@ -266,6 +266,22 @@ async def guild_permissions(self) -> Permissions:
...
 
+    async def has_permission(self, *permissions: Permissions) -> bool:
+        """
+        Checks if the member has all the given permission(s)
+        Args:
+            permissions: The permission(s) to check whether the user has it.
+        """
+
+        # Get the user's permissions
+        guild_permissions = self.guild_permissions()
```
benwoo1110: `self.guild_permissions()` is an async method. ~~I also think self.guild_permissions() method itself is broken 👀.~~

```suggestion
        guild_permissions = await self.guild_permissions()
```

dis_snek/models/discord_objects/user.py
```diff
@@ -266,6 +266,34 @@ async def guild_permissions(self) -> Permissions:
 
         return permissions
 
+    async def has_permission(self, *permissions: Permissions) -> bool:
```
benwoo1110, line 269: hmm not too sure if we should have an *arg parameter since its bitwise and we could just do e.g. `admin.has_permission(Permissions.KICK_MEMBERS | Permissions.BAN_MEMBERS)`

It Will be slightly more efficient without for loop before, but perhaps some that don't know bitwise stuff will be confused... @LordOfPolls 😅

benwoo1110 (commented):
Are you sure you tested the code? Doesn't seem to work for me, also left some other comments.

dis_snek/models/discord_objects/user.py
Kigstn: Both should work at the moment. But the bitwise flag way is giving me type checker errors which I have no idea how to fix

```
await has_permission(Permissions.KICK_MEMBERS, Permissions.CREATE_INSTANT_INVITE)
> True
await has_permission(Permissions.ADMINISTRATOR, Permissions.CREATE_INSTANT_INVITE)
> False

await has_permission(Permissions.KICK_MEMBERS | Permissions.CREATE_INSTANT_INVITE)
> True
await has_permission(Permissions.ADMINISTRATOR | Permissions.CREATE_INSTANT_INVITE)
> False
```

LordOfPolls:
By fixing the `guild_permissions` bug in your code, you've actually broken this PR, as the member object currently does not have a functioning guild reference, that said, **your** code should work fine. So i'm happy to merge

LordOfPolls:
[`member.has_permission` - Check if member has perm(s)](https://trello.com/c/cUmdXpGb/124-memberhaspermission-check-if-member-has-perms)

PR #38, opened 2021-10-19
Fix ctx.edit_origin() and message.edit()

Kigstn:

- [x] Breaking code change

## Description

ctx.edit_origin() and message.edit() had different arguments than .send()

I changed them to function the same, doesn't make sense to have `file` on .send() and `filepath` on .edit() for example.

This is technically breaking since an argument changed

## Changes

- ctx.edit_origin() and message.edit() now have the same args as .send()

## Checklist

- [ ] I've tested my code


benwoo1110 (approved):
(I totally wasn't the one that create this bug)

PR #46, opened 2021-10-20
on_message_reaction listeners

Wolfhound905:

- [ ] Breaking code change

## Description

Adds listeners for reaction events. Example usage.
```py
@listen()
async def on_message_reaction_remove_all(event):
    print(event.message.content)

@listen()
async def on_message_reaction_remove(event):
    print(event.emoji.name)

@listen()
async def on_message_reaction_add(event):
    print(event.emoji.name)
```
add and remove returns the CustomEmoji class.
remove_all returns message class

## Changes

- Adds reaction events

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client.py
+++ b/dis_snek/client.py
@@ -52,6 +52,7 @@ from dis_snek.models import (
     MessageContext,
     AutocompleteContext,
     ComponentCommand,
+    CustomEmoji,
 )
 from dis_snek.models.enums import ComponentTypes, Intents, InteractionTypes, Status, ActivityType
 from dis_snek.models.events import RawGatewayEvent, MessageCreate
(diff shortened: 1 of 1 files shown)


dis_snek/client.py
```diff
@@ -825,6 +826,52 @@ async def _on_raw_message_update(self, event: RawGatewayEvent) -> None:
...
+            event: raw reaction add event
+        """
+        if member := event.data.get("member"):
+            author = self.cache.place_member_data(event.data.get("guild_id"), member)
+        else:
+            author = await self.cache.get_user(event.data.get("user_id"))
+
+        emoji = CustomEmoji.from_dict(event.data.get("emoji"), self)
+
+        self.dispatch(events.MessageReactionAdd(message=None, emoji=emoji, author=author))
```
benwoo1110, line 844: Why is `message=None` here? I think either we change to only pass the `message_id` or get the full message from the cache.

dis_snek/client.py
```diff
@@ -825,6 +826,52 @@ async def _on_raw_message_update(self, event: RawGatewayEvent) -> None:
...
+            event: raw reaction remove event
+        """
+        if member := event.data.get("member"):
+            author = self.cache.place_member_data(event.data.get("guild_id"), member)
+        else:
+            author = await self.cache.get_user(event.data.get("user_id"))
+
+        emoji = CustomEmoji.from_dict(event.data.get("emoji"), self)
+
+        self.dispatch(events.MessageReactionRemove(message=None, emoji=emoji, author=author))
```
benwoo1110, line 861: Same comment as above

dis_snek/client.py
```diff
@@ -825,6 +826,52 @@ async def _on_raw_message_update(self, event: RawGatewayEvent) -> None:
...
+
+        Args:
+            event: raw reaction remove event
+        """
+        if member := event.data.get("member"):
+            author = self.cache.place_member_data(event.data.get("guild_id"), member)
+        else:
+            author = await self.cache.get_user(event.data.get("user_id"))
+
+        emoji = CustomEmoji.from_dict(event.data.get("emoji"), self)
```
benwoo1110, line 859: The gateway only returns partial emoji data, I think should use the `Emoji` class instead.

benwoo1110 (changes requested):
Left some comments. Also, have you tested them to be working?

dis_snek/client.py
Wolfhound905: I have not tried sending just the message_id. I can at another time.

PR #55, opened 2021-10-25
Remove `attr_utils.define` shortcut

AstreaTSS:

- [ ] Breaking code change

## Description

A shortcut for `attr.define` with commonly-used options (`attr_utils.define`) had been used for a decent bit of the code. However, it also broke Visual Studio Code's typehinting, with the typehinting reporting the classes defined with this as `None` instead of themselves. This PR removes the shortcut in order to fix the typehinting and replaces the current instances of it with what the shortcut did.

## Changes

- Removed `attr_utils.define`
- Removed instances where `attr_utils.define` was used, and replaced it with what it did behind the scenes.
  - Special cases where parts of the shortcut were changed around were accounted for.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/context.py
+++ b/dis_snek/models/context.py
@@ -71,7 +71,14 @@ class Resolved:
         return new_cls
 
 
-@define
+@attr.define(
+    eq=False,
+    order=False,
+    hash=False,
+    slots=True,
+    kw_only=True,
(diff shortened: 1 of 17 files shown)


AstreaTSS:
Closing in case there's a better solution because there might.

PR #56, opened 2021-10-25
Fix attr_utils.define shortcut for typehinting

AstreaTSS:

- [ ] Breaking code change

## Description

A shortcut for `attr.define` with commonly-used options (`attr_utils.define`) had been used for a decent bit of the code. However, it also broke Visual Studio Code's typehinting, with the typehinting reporting the classes defined with this as `None` instead of themselves. This PR fixes that.

## Changes

- Fix attr_utils.define shortcut for typehinting by using TypeVar and Callable

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/utils/attr_utils.py
+++ b/dis_snek/utils/attr_utils.py
@@ -1,6 +1,6 @@
 import logging
 from functools import partial
-from typing import Any, Dict
+from typing import Any, Dict, TypeVar, Callable
 
 import attr
 from dis_snek.const import logger_name, MISSING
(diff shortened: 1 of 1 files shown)


benwoo1110:
What about type hinting for `field`?

AstreaTSS:
> What about type hinting for `field`?

Surprisingly enough, it *isn't* needed. Works well enough.
I *think* there's a note in `attrs`'s source code that whatever issue is going on only affects `define` and something else, but yeah, doesn't affect `field`.

PR #59, opened 2021-10-28
INVISIBLE on bot stop

Wolfhound905:

- [ ] Breaking code change

## Description

Fixes bots appearing online for some time after `bot.stop()` or <kbd>Ctrl</kbd> + <kbd>c</kbd>

## Changes

- Set the bot to INVISIBLE when going offline.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client.py
+++ b/dis_snek/client.py
@@ -398,6 +398,7 @@ class Snake(
 
     async def stop(self):
         log.debug("Stopping the bot.")
+        await self.change_presence(Status.INVISIBLE)
         await self.ws.close()
 
     def dispatch(self, event: events.BaseEvent, *args, **kwargs):


LordOfPolls:
This won't be merged. It burns an api call, when in reality the bot should just tell the API it's logging off

Wolfhound905:
Would telling the API you're logging off make the status go to offline. Or will the bot always stay "online" for a few minutes after being shut off.

LordOfPolls:
No, you're changing the presence. The bot stays online as we don't shut down the bot properly, as of yet. We're just closing the websocket

This just patches over that, when in reality we need to implement the proper closing method

Edit: re your edit, yes the bot would be offline if implimented properly

PR #63, opened 2021-10-30
Add `wait_for_component()`

Toricane:

- [ ] Breaking code change

## Description

This PR adds `Snake.wait_for_component()`
How to use:
```py
...
# Example check:
def check(event: Component):
    ctx: ComponentContext = event.context
    return str(ctx.author.id) == ctx.custom_id


message = await ctx.send(..., components=your_components)
component_event = await bot.wait_for_component(messages=message, components=your_components, check=check)
component_ctx = component_event.context
# all kwargs are optional
```

## Changes

- Added `wait_for_component()`
- ~~Added `get_messages_ids()`~~ Removed and used `to_snowflake_list()` instead
## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client.py
+++ b/dis_snek/client.py
@@ -5,7 +5,7 @@ import inspect
 import logging
 import sys
 import traceback
-from typing import TYPE_CHECKING, Callable, Coroutine, Dict, List, Optional, Union, Awaitable
+from typing import TYPE_CHECKING, Callable, Coroutine, Dict, List, Optional, Union
 
 import aiohttp
 
@@ -45,6 +45,7 @@ from dis_snek.models import (
     MessageCommand,
     BaseCommand,
     to_snowflake,
+    to_snowflake_list,
     ComponentContext,
     InteractionContext,
     MessageContext,
@@ -57,7 +58,7 @@ from dis_snek.models.enums import ComponentTypes, Intents, InteractionTypes, Sta
 from dis_snek.models.events import RawGatewayEvent, MessageCreate
 from dis_snek.models.events.internal import Component
 from dis_snek.models.wait import Wait
-from dis_snek.models.discord_objects.components import get_components_ids, get_messages_ids
+from dis_snek.models.discord_objects.components import get_components_ids, BaseComponent
 from dis_snek.smart_cache import GlobalCache
 from dis_snek.utils.cache import TTLCache
 from dis_snek.utils.input_utils import get_first_word
@@ -467,10 +468,12 @@ class Snake(
     async def wait_for_component(
         self,
         messages: Union[Message, int, list] = None,
-        components: Union[str, dict, list] = None,
+        components: Optional[
+            Union[List[List[Union["BaseComponent", dict]]], List[Union["BaseComponent", dict]], "BaseComponent", dict]
+        ] = None,
         check=None,
         timeout=None,
-    ) -> Component:
+    ):
         """
         Waits for a message to be sent to the bot.
 
@@ -486,7 +489,7 @@ class Snake(
         if not (messages or components):
             raise ValueError("You must specify messages or components (or both)")
 
-        message_ids = list(get_messages_ids(messages)) if messages else None
+        message_ids = to_snowflake_list(messages) if messages else None
         custom_ids = list(get_components_ids(components)) if components else None
 
         # automatically convert improper custom_ids
@@ -495,12 +498,14 @@ class Snake(
 
         def _check(event: Component):
             ctx: ComponentContext = event.context
-            if check and not check(event):
-                return False
             # if custom_ids is empty or there is a match
             wanted_message = not message_ids or ctx.origin_message_id in message_ids
             wanted_component = not custom_ids or ctx.custom_id in custom_ids
-            return wanted_message and wanted_component
+            if wanted_message and wanted_component:
+                if check is None or check(event):
+                    return True
+                return False
+            return False
 
         return await self.wait_for("component", checks=_check, timeout=timeout)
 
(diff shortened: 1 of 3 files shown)


dis_snek/utils/attr_utils.py
```diff
@@ -25,8 +25,7 @@
 _T = TypeVar("_T")
 
 
-def define(cls: _T) -> _T:
-    return attr.define(cls, **class_defaults)
+define: Callable[[_T], _T] = partial(attr.define, **class_defaults)
```
benwoo1110: Why is this part of the pr? Doesnt look like its used for `wait_for_component`

dis_snek/models/discord_objects/components.py
```diff
@@ -382,6 +384,22 @@ def get_components_ids(component: Union[str, dict, list, InteractiveComponent])
         raise ValueError(f"Unknown component type of {component} ({type(component)}). " f"Expected str, dict or list")
 
 
+def get_messages_ids(message: Union[int, "Message", list]) -> Iterator[int]:
```
benwoo1110: I feel like we can just use the generic `to_snowflake_list()`?

dis_snek/client.py
```diff
@@ -462,6 +464,46 @@ def wait_for(self, event: str, checks: Optional[Callable[..., bool]] = MISSING,
...
+        message_ids = list(get_messages_ids(messages)) if messages else None
+        custom_ids = list(get_components_ids(components)) if components else None
+
+        # automatically convert improper custom_ids
+        if custom_ids and not all(isinstance(x, str) for x in custom_ids):
+            custom_ids = [str(i) for i in custom_ids]
+
+        def _check(event: Component):
+            ctx: ComponentContext = event.context
+            if check and not check(event):
```
benwoo1110: I think custom check callback should ran after checking that message and custom_id matches.

dis_snek/client.py
```diff
@@ -462,6 +464,46 @@ def wait_for(self, event: str, checks: Optional[Callable[..., bool]] = MISSING,
 
         return asyncio.wait_for(future, timeout)
 
+    async def wait_for_component(
+        self,
+        messages: Union[Message, int, list] = None,
+        components: Union[str, dict, list] = None,
+        check=None,
+        timeout=None,
+    ) -> Component:
```
benwoo1110: it returns an awaitable future, not component object.

benwoo1110 (changes requested):

dis_snek/utils/attr_utils.py
Toricane: Check #general and #lib-discussion, Proxy, Astrea and I were talking about this, it's there because #62 wasn't properly tested and basically broke the lib, so we decided to revert

benwoo1110: Yes ik, but you shouldn't do it in this PR. It's completely not related to `wait_for_component()`

AstreaTSS: Should be noted that #64 should *properly* fix this bug. But yeah, keep this for local testing (until #64 is merged) but *not* for this PR.

Toricane: Got it, thanks!

dis_snek/models/discord_objects/components.py
Toricane: How do I do that?

dis_snek/utils/attr_utils.py
Toricane: Reverted this and upstreamed

dis_snek/models/discord_objects/components.py
Toricane: So instead of
```py
message_ids = list(get_messages_ids(messages)) if messages else None
```
I should do
```py
message_ids = to_snowflake_list(messages) if messages else None
```
? Or is this wrong?

Toricane: Done this, time to test it

dis_snek/client.py
Toricane: did this instead, think thats what you wanted:
```py
def _check(event: Component):
    ctx: ComponentContext = event.context
    # if custom_ids is empty or there is a match
    wanted_message = not message_ids or ctx.origin_message_id in message_ids
    wanted_component = not custom_ids or ctx.custom_id in custom_ids
    if wanted_message and wanted_component:
        if check is None or check(event):
            return True
        return False
    return False
```

dis_snek/client.py
Toricane: How do I typehint that? `Awaitable[Future]` doesn't work since Future isn't in typing, so I just removed the typehint for now

dis_snek/client.py
Toricane: this worked:
```py
def _check(event: Component):
    ctx: ComponentContext = event.context
    # if custom_ids is empty or there is a match
    wanted_message = not message_ids or ctx.message.id in (
        [message_ids] if isinstance(message_ids, int) else message_ids
    )
    wanted_component = not custom_ids or ctx.custom_id in custom_ids
    if wanted_message and wanted_component:
        if check is None or check(event):
            return True
        return False
    return False
```

dis_snek/models/discord_objects/components.py
Toricane: Oops, I forgot to reply, works

dis_snek/client.py
Toricane: Nevermind, I added `from asyncio import Future` into `if TYPE_CHECKING:` and then used typehint `Awaitable['Future']`

benwoo1110 (approved):
I think looks ok to me. @LordOfPolls what do you think?

Wolfhound905:
I am going to do some testing.

PR #64, opened 2021-10-31
Actual fix to `attr_utils.define`

AstreaTSS:

- [ ] Breaking code change

## Description

Actually, ACTUALLY fix `attr_utils.define`. No more typehinting nonsense. No more errors with positional arguments or whatever. An *actual* fix.
On a more serious note, this fix follows what `attr` did to fix its own `define` function and what it recommends while extending it... like we are here.
Link: [here](https://www.attrs.org/en/stable/extending.html).

## Changes

- Revert `attr_utils.py` back to how it was before any of my 'fixes'.
- Created a `attr_utils.pyi`, which has the typehinting needed for `define`.

## Checklist

- [x] I've tested my code (checked pretty much everything I could, so *hopefully* this works)


AstreaTSS:
If possible, I would like someone to review these changes and test them out so we don't have to keep doing this over and over again.

LordOfPolls:
Seems to work, merging

PR #69, opened 2021-11-04
annotations

LordOfPolls:

- [ ] Breaking code change

## Description

This PR adds the annotation system that was discussed in #libdiscussion of the Discord Server

## Changes

- Add `command.call_callback` which handles parameter parsing 
- Add default annotations to `dis_snek/annotations/argument_annotations.py`
- Add `context.contet_parameters`
- Add `utils/get_parameters`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/annotations/argument_annotations.py
+++ b/dis_snek/annotations/argument_annotations.py
@@ -27,7 +27,7 @@ def define_annotation():
         params = get_parameters(func)
         args = []
         for param in params.values():
-            if param.annotation is Context:
+            if issubclass(param.annotation, Context):
                 args.append("context")
             elif param.annotation is Scale:
                 args.append("scale")
--- a/dis_snek/annotations/argument_annotations.py
+++ b/dis_snek/annotations/argument_annotations.py
@@ -1,0 +1,68 @@
+from typing import Union, TYPE_CHECKING, Any, List, Type, Dict
+
+from dis_snek.const import MISSING
+from dis_snek.utils.misc_utils import get_parameters
+
+from dis_snek.models.context import Context
+from dis_snek.models.scale import Scale
+from dis_snek.client import Snake
+
+if TYPE_CHECKING:
+    from dis_snek.models import Member, User, TYPE_MESSAGEABLE_CHANNEL
+
+
+def define_annotation():
+    """
+    Define a function as an annotation.
+
+    The function **must** type hint its expected arguments. Make sure your annotation runs quickly,
+    as command execution will be delayed until all annotations are completed
+
+    **Supported Types:**
+    `Context`
+    `Scale`
+    """
+
+    def wrapper(func):
+        params = get_parameters(func)
+        args = []
+        for param in params.values():
+            if param.annotation is Context:
+                args.append("context")
+            elif param.annotation is Scale:
+                args.append("scale")
+            # elif param.annotation is Snake:
+            #     args.append("snake")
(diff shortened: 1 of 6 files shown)


LordOfPolls:
This PR does not add the slash command variant of annotations

Wolfhound905:
> 69

nice. 😎

dis_snek/utils/misc_utils.py
```diff
@@ -74,3 +74,10 @@ def wrap_partial(obj, cls):
         obj.subcommands = {k: wrap_partial(v, cls) for k, v in obj.subcommands.items()}
 
     return obj
+
+
+def get_parameters(callback: Callable):
```
artem30801, line 79: Maybe dict comprehension would be better/faster

dis_snek/models/command.py
```diff
@@ -98,6 +108,47 @@ async def __call__(self, context, *args, **kwargs):
...
+            return await callback()
+
+        c_args = copy.copy(context.args)
+        for param in parameters.values():
+            if config := getattr(param.annotation, "_annotation_dat", None):
+                # if user has used an snek-annotation, run the annotation, and pass the result to the user
+                local = {"context": context, "scale": self.scale}
+                ano_args = [local[c] for c in config["args"]]
+                if param.kind != param.POSITIONAL_ONLY:
+                    kwargs[param.name] = param.annotation(*ano_args)
```
artem30801, line 127: No async anno function support?

dis_snek/annotations/argument_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+def define_annotation():
+    """
+    Define a function as an annotation.
+
+    The function **must** type hint its expected arguments. Make sure your annotation runs quickly,
+    as command execution will be delayed until all annotations are completed
+
+    **Supported Types:**
+    `Context`
+    `Scale`
```
artem30801, line 23: No `Snake` support?

dis_snek/annotations/argument_annotations.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    def wrapper(func):
+        params = get_parameters(func)
+        args = []
+        for param in params.values():
+            if param.annotation is Context:
```
artem30801: I think it would be better to check with `issubclass()`, so users can use `InteractionContext` etc for their type hinting

dis_snek/annotations/argument_annotations.py
LordOfPolls: No, remember our conversation in #lib-discussion? Commands dont have a reference to client

dis_snek/annotations/argument_annotations.py
LordOfPolls: Will be resolved in a follow-up commit

dis_snek/models/command.py
LordOfPolls: Calling a coroutine is slower than calling a regular function, baring in mind this code would be used by application commands, it seems like a *very* bad idea to allow for coroutines, which could potentitally be blocked by the event loop for an indeterminate amount of time, causing interactions to fail

dis_snek/utils/misc_utils.py
LordOfPolls: I'd prefer to leave it in its own method ( i intend to reuse this), but ill convert it to use comprehension instead

PR #77, opened 2021-11-17
Allow value to be Any type

compil3:
Allows Embed value to be Any type as to avoid having to type cast each value.

- [ ] Breaking code change

## Description
Currently, if Embed values are anything besides a str unless you type cast the value, it will fail. Example:

>e.add_field("test", 1)
>TypeError: object of type 'int' has no len()  

Discord requires `value` to be `str`; changing value to Any resolves the issue and idiot proofs it.

## Changes
-  From `def add_field(self, name: str, value: str, inline: bool = False) -> None:`  to  `def add_field(self, name: str, value: Any, inline: bool = False) -> None:`
- From `self.fields.append(EmbedField(name, value, inline))` to `self.fields.append(EmbedField(name, str(value), inline))`

## Checklist

- [X] I've ensured my code works on `Python 3.9.x`
- [X] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/embed.py
+++ b/dis_snek/models/discord_objects/embed.py
@@ -253,7 +253,7 @@ class Embed(DictSerializationMixin):
         """
         self.footer = EmbedFooter(text=text, icon_url=icon_url)
 
-    def add_field(self, name: str, value: str, inline: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


compil3:
1st PR i've ever done, let me know if I'm an idiot.

PR #81, opened 2021-11-18
💥 Event arg dispatching changes

LordOfPolls:

- [x] Breaking code change

## Description

This PR changes how event arguments are dispatched to events. Instead of passing the event object itself, this PR grabs the attributes from the event, and passes those to the listener, either positionally, or through keywords. 

## Changes

- Modified `dispatch` to grab attributes from event object
- Modified `_queue_task` to check params
  - This determines whethere kwargs or args should be used
- Pass args/kwargs to listeners instead of `Event` object
- Modified `event_processors` to use this new paradigm

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client.py
+++ b/dis_snek/client.py
@@ -366,7 +366,6 @@ class Snake(
             except asyncio.CancelledError:
                 pass
             except Exception as e:
-                breakpoint()
                 await self.on_error(event, e)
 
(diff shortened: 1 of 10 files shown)


LordOfPolls:
Fun fact, this method is so intuitive, @Kigstn wrote the docs assuming this would be how we dispatched event arguments

PR #92, opened 2021-11-28
Fix thread Unions, Add an await

Wolfhound905:

- [ ] Breaking code change

## Description

I was using more with Discord threads and ran into a few issues and here were my fixes.

## Changes

- Add an await
- Convert to Optional instead of Union

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/channel.py
+++ b/dis_snek/models/discord_objects/channel.py
@@ -415,7 +415,7 @@ class ThreadableMixin:
         )
         return self._client.cache.place_channel_data(thread_data)
 
-    async def get_public_archived_threads(self, limit: int = None, before: Union["Timestamp"] = None) -> ThreadList:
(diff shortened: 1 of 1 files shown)


dis_snek/models/discord_objects/channel.py
```diff
@@ -415,7 +415,9 @@ async def create_thread_without_message(
         )
         return self._client.cache.place_channel_data(thread_data)
 
-    async def get_public_archived_threads(self, limit: int = None, before: Union["Timestamp"] = None) -> ThreadList:
+    async def get_public_archived_threads(
+        self, limit: int = None, before: Union["Timestamp", None] = None
```
silasary: This can be Optional["Timestamp"]

Wolfhound905: Ah ya true

PR #93, opened 2021-11-29
If slash_command(description=) isn't set, attempt to pull it from the coroutine's docstring

silasary:

- [ ] Breaking code change

## Description
I find putting ~80 character strings in the decorator is worse on code readability than using a docstring to explain what a command does. 

## Changes
This PR adds support for checking the docstring before falling back to "No Description Set"

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/application_commands.py
+++ b/dis_snek/models/application_commands.py
@@ -463,7 +463,7 @@ class ComponentCommand(InteractionCommand):
 
 def slash_command(
     name: str,
-    description: str = "No Description Set",
+    description: str = MISSING,
     scopes: List["Snowflake_Type"] = MISSING,
     options: Optional[List[Union[SlashCommandOption, Dict]]] = None,
     default_permission: bool = True,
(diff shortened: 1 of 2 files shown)


philip-ulrich:
This is pretty smooth. I like this idea

Wolfhound905:
How would you use it?

philip-ulrich:
You add the description to your functions docstring.

```python
@slash_command(
    name="info",
    group_name="mods",
)
async def info_mods(self, ctx: Context):
    """Shows information about mods"""
    await ctx.send(en_US.call_mods)
```

today, it looks like this:

```python
@slash_command(
    name="info",
    group_name="mods",
    description="Shows information about mods",
)
async def info_mods(self, ctx: Context):
    await ctx.send(en_US.call_mods)
```

This PR makes either of those valid and decorator takes precedence over docstring

Wolfhound905:
Oh that's neat, I like the idea too.

PR #94, opened 2021-12-02
Add `Role.members` and make `Member.has_role` synchronous

Catalyst4222:

- [x] Breaking code change

## Description

Turn `Member.has_role` into a synchronous function, and add the `Role.members` property
This pr is breaking because devs who used `Member.has_role` will have to remove the preceding `await`. Because of that, this pr should probably only be merged once a new major version is about to be published

## Changes

- Turned Member.has_role into a synchronous function
- Changed all uses of the above to reflect changes (except for the one that wasn't awaited)
- add the `Role.members` property, which iterates over the members in a guild to check if they have a role

## Checklist

- [x] I've tested my code


LordOfPolls:
Pypi versioning has no impact on what's on github. The entire github build is breaking compared to Pypi as we're staging multiple breaking changes. As such this will be merged now.

Thanks for the contribution

PR #100, opened 2021-12-04
Fix `ComponentContext.values`

Catalyst4222:

- [ ] Breaking code change

## Description

Previously, `ComponentContext.values` would always give an empty list. Now it gives a list of the values of the options the user chose (in string form) 

## Changes

- `ComponentContext.from_dict` sets `.values` to `data['data']['values']`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/context.py
+++ b/dis_snek/models/context.py
@@ -251,7 +251,11 @@ class ComponentContext(InteractionContext):
         new_cls.custom_id = data["data"]["custom_id"]
         new_cls.component_type = data["data"]["component_type"]
         new_cls.message = client.cache.place_message_data(data["message"])
-        new_cls.values = data["data"]["values"]
+
+        if "values" in data["data"]:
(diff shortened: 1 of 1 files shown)


dis_snek/models/context.py
```diff
@@ -252,6 +252,11 @@ def from_dict(cls, data: Dict, client: "Snake") -> "ComponentContext":
         new_cls.component_type = data["data"]["component_type"]
         new_cls.message = client.cache.place_message_data(data["message"])
 
+        if "values" in data["data"]:
+            new_cls.values = data["data"]["values"]
+        else:
+            new_cls.values = []
```
LordOfPolls: You could have just used `data['data'].get("values", [])`

Catalyst4222: 😅 
Should I try and revert it or just add another commit?

silasary: Just add another commit

PR #106, opened 2021-12-14
Add VoiceStateUpdate event

Kigstn:

- [ ] Breaking code change

## Description

This adds the VoiceStateUpdate event and all required classes. 

This is the first time I dove deep into sneks codebase, so I probly did something suboptiomal. I would appreciate a double check, just in case

## Changes

- add VoiceStateUpdate event
- add VoiceState model
- cache VoiceState

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/voice_state.py
+++ b/dis_snek/models/discord_objects/voice_state.py
@@ -24,9 +24,9 @@ class VoiceState(ClientObject):
     mute: bool = field(default=False)
     self_deaf: bool = field(default=False)
     self_mute: bool = field(default=False)
-    self_stream: Optional[bool] = field(default=None)
-    self_video: bool = field(default=MISSING)
-    suppress: bool = field(default=MISSING)
+    self_stream: Optional[bool] = field(default=False)
+    self_video: bool = field(default=False)
+    suppress: bool = field(default=False)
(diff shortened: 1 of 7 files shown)


dis_snek/models/discord_objects/voice_state.py
```diff
@@ -0,0 +1,58 @@
...
+class VoiceState(ClientObject):
+    user_id: "Snowflake_Type" = field(default=MISSING, converter=to_snowflake)
+    session_id: str = field(default=MISSING)
+    deaf: bool = field(default=False)
+    mute: bool = field(default=False)
+    self_deaf: bool = field(default=False)
+    self_mute: bool = field(default=False)
+    self_stream: Optional[bool] = field(default=None)
+    self_video: bool = field(default=MISSING)
+    suppress: bool = field(default=MISSING)
```
artem30801: IMO type hinting should include MISSING as union with bool or sane bool defaults should be assumed

Kigstn: Good catch, missed that!

LordOfPolls:
I had no intention of adding voice support for now, but this seems like a good core implimentation, so it'll be merged

PR #112, opened 2021-12-17
Add Scheduled Events support

Wolfhound905:

- [ ] Breaking code change

## Description

Added support for Discord's new Scheduled Events. 
Also, I am creating this draft because I would like some people to try this and tell me what could be improved. Feedback would be appreciated 👍

Examples: 
- [create event slash command](https://gist.github.com/Wolfhound905/662379da5963c83b41b29db054db4d54) 
- [bot.py](https://gist.github.com/Wolfhound905/e5810156b9ded86c1583a14dc7d34633)

## Changes

- Add Scheduled Events Support

## Checklist

- [x] I've tested my code


silasary (approved):
I've tested creating and listing guild events, and it feels good

dis_snek/models/discord_objects/invite.py
```diff
@@ -66,6 +66,9 @@ def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
         if "target_application" in data:
             data["target_application"] = Application.from_dict(data, client)
 
+        if "target_event_id" in data:
+            data["scheduled_event"] = data["target_event_id"]
+
         if channel := data.pop("channel", None):
```
i0bs, line 72: Why are you checking for an assignment expression when it will already provide `None` for the actual channel data?

Wolfhound905: I am confused about what you are saying. Could you explain in the discord?

dis_snek/client.py
```diff
@@ -1392,6 +1393,21 @@ async def get_member(self, user_id: "Snowflake_Type", guild_id: "Snowflake_Type"
         """
         return await self.cache.get_member(guild_id, user_id)
 
+    async def get_scheduled_event(
+        self, guild_id: "Snowflake_Type", scheduled_event_id: "Snowflake_Type", with_user_count: bool = False
+    ) -> "ScheduledEvent":
```
LordOfPolls, line 1386: This doesn't actually *need* to be in quotations, you *need* to use quoations when your type hint is imported through `TYPE_CHECKING` or if youre forward referencing 
(don't act on this, just mark as resolved)

dis_snek/models/discord_objects/invite.py
Wolfhound905: Oh wait, you mean I should do something like this?
```py
if data["target_event_id"]:
    data["scheduled_event"] = data["target_event_id"]
```

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # Discord only has one at the momment.
+    SCHEDULED = 1
+    ACTIVE = 2
+    COMPLETED = 3
+    CANCELED = 4
+
+
+@define()
+class ScheduledEventUser(DiscordObject):
```
LordOfPolls: Why is this object necessary? Afiak there is no extra information over a standard `member` object when a member is part of an event?

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+    user: "User" = attr.ib()
+    member: "Member" = attr.ib(default=MISSING)
```
LordOfPolls: Why are you seperating user and member? 
`ScheduledEvent` is short for `GuildScheduledEvent`, they always occur within a guild so this *should* always be a member.

LordOfPolls: Also member contains a user object

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+        data["id"] = data["user"].id
+        return data
+
+
+@define()
+class ScheduledEvent(DiscordObject):
+    name: str = attr.ib()
+    description: str = attr.ib(default=MISSING)
+    entity_type: Union[ScheduledEventEntityType, int] = attr.ib(converter=ScheduledEventEntityType)
+    """	the type of the scheduled event"""
```
LordOfPolls: Whats with the indent in this, and a few other, docstrings?

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+        return await self._client.cache.get_user(self._creator_id) if self._creator_id else None
+
+    @property
+    def guild(self) -> "Guild":
+        return self._client.cache.guild_cache.get(self._guild_id)
+
+    @classmethod
+    def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
+        if data.get("creator"):
+            data["creator"] = client.cache.place_user_data(data["creator"])
```
LordOfPolls, line 101: `creator` is a user in the api spec, but you could *attempt* to get a member object from the cache as this **will** occur in a guild

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+        self,
+        limit: Optional[int] = 100,
+        with_member_data: bool = False,
+        before: Optional["Snowflake_Type"] = MISSING,
+        after: Optional["Snowflake_Type"] = MISSING,
+    ) -> Optional[List["ScheduledEventUser"]]:
+        """Get event users
+
+        Args:
+            limit: Discord defualts to 100
```
LordOfPolls, line 141: These args dont match the signature of this method

dis_snek/models/discord_objects/scheduled_event.py
```diff
@@ -0,0 +1,204 @@
...
+        event_users = await self._client.http.get_scheduled_event_users(
+            self._guild_id, self.id, limit, with_member_data
+        )
+        if event_users:
+            return ScheduledEventUser.from_list(event_users, self._client)
+        else:
+            return None
+
+    async def delete(self, reason: str = MISSING) -> None:
+        """Deletes this event"""
```
LordOfPolls: this method has arguments, you need to add a `args` section to the docstring

LordOfPolls (changes requested):

dis_snek/models/discord_objects/scheduled_event.py
Wolfhound905: Could you help me out? I am not sure what else I would do. I did this because I just saw how the team.py file did it.

dis_snek/client.py
Wolfhound905: Thank's for the info.

dis_snek/models/discord_objects/scheduled_event.py
Wolfhound905: I just saw that member could not be there sometimes 
https://discord.com/developers/docs/resources/guild-scheduled-event#get-guild-scheduled-event-users
If you could help with this one that would be nice, thanks.

Wolfhound905: image

i0bs: https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-user-object

`user` and `member` should remain separate as per the API schema.

Wolfhound905: So is my method good? Or still need to be changed?

silasary: I'd keep it as is

LordOfPolls: I disagree with the above. One this object shouldn't even exist.

And two, Events ought to have an attribute called `members` or `participants` or whatever name which is a `List[Union[User, Member] `

That'd be the solution with the best UX

dis_snek/models/discord_objects/scheduled_event.py
LordOfPolls: Teams.py uses this paradigm because it adds additional information to the objects. Here you're adding nothing so it's a redundant object

Easiest solution would be to store the list of participant snowflakes, then make a property which returns the list of members using a generator

LordOfPolls: Tho, Events could easily have thousands of participants... Maybe an async iterator?

Wolfhound905: Sorry, I don't think I am comfortable adding something like this yet. I am gonna have to pass this on to someone else and watch what they do so I can see how this would be done

dis_snek/models/discord_objects/scheduled_event.py
i0bs: > I disagree with the above. One this object shouldn't even exist.
> 
> And two, Events ought to have an attribute called `members` or `participants` or whatever name which is a `List[Union[User, Member] `
> 
> That'd be the solution with the best UX

I think the attributes existence should be respected if you're trying to keep it strict to the schema design, since a majority of every other object I've thus far seen does. From a UX perspective you're right, but if its for a developer transitioning from one library to another, I'm worried it might add on to confusion.

dis_snek/models/discord_objects/invite.py
```diff
@@ -81,4 +84,6 @@ def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
 
     @property
     def link(self) -> str:
+        if self.scheduled_event:
+            return f"https://discord.gg/{self.code}?event={self.scheduled_event}"
         return f"https://discord.gg/{self.code}"
```
i0bs, line 89:
```py
        return (
            f"https://discord.gg/{self.code}" if self.scheduled_event
            else f"https://discord.gg/{self.code}"
        )
```

This isn't really *required* per-se, but it would be more consistent with how black generalizes formatting for ternary returns.

dis_snek/models/discord_objects/guild.py
```diff
@@ -662,6 +664,80 @@ async def delete_channel(self, channel: Union["TYPE_GUILD_CHANNEL", "Snowflake_T
...
+            channel_id: channel id
+            entity_metadata: event metadata
+            privacy_level: event privacy level
+            reason: reason for creating this scheduled event
+
+        Returns:
+            ScheduledEvent object
+
+        !!! note
+            channel_id is Optional when entity_type is `ScheduledEventEntityType.EXTERNAL`
```
i0bs:
```suggestion
            channel_id is Optional when entity_type is `ScheduledEventEntityType.EXTERNAL`
            
            location must also be supplied when this is true.
```

dis_snek/models/discord_objects/scheduled_event.py
i0bs: From what I'm understanding, you want something like this?
( this is pseudocode, me no dis-snek :( )

```py
    ...
    from ... import User as EventUser # name this whatever, we're trying to avoid the notorious circular import issue.

    @property
    async def users(self):
        _users: list = [EventUser(user) for user in self.participants]
```

dis_snek/models/discord_objects/scheduled_event.py
LordOfPolls: We're not strictly following the discord dev schema, UX takes priority imo. I follow the docs where it doesnt harm UX

dis_snek/models/discord_objects/guild.py
silasary: Yep. Location should definitely be a first-class parameter here, rather than something hidden in entity_metadata

dis_snek/models/discord_objects/invite.py
silasary: I personally feel like the current way is more readable.  No need to turn it into a ternary.

dis_snek/models/discord_objects/guild.py
Wolfhound905: This is just in case entity_metadata receives more things in the future.

silasary: I'm not saying to remove entity_metadata.  Just allow users to specify location without it

dis_snek/models/discord_objects/invite.py
Wolfhound905: Ya I am going to agree with @silasary. The current method is just better to look at.

LordOfPolls:
Any movement on this PR?

Wolfhound905:
> Any movement on this PR?

image

silasary:
https://github.com/Wolfhound905/Dis-Snek/pull/1

dis_snek/models/discord_objects/scheduled_event.py
i0bs: this is to help out wolfhound here: is that a request or a thought you jotted down?

dis_snek/models/discord_objects/scheduled_event.py
i0bs: Alright, I dig the thought. My only suggestion is maybe considering a property that can access that part of the API schema without actually having an attribute dedicated to it, so you get the best of both worlds and you don't end up *actually* writing a value for an attr you have to handle.

LordOfPolls: I don't follow. How will a property provide access to the api-style schema?

dis_snek/models/discord_objects/scheduled_event.py
LordOfPolls: A request

dis_snek/models/discord_objects/scheduled_event.py
i0bs: Not sure how you do it with your serializer mixin but something like this perhaps

```py
@property
def member(self):
    _member: "Member"
    return _member = attr.ib()
```

LordOfPolls: I don't see what benifit that would offer? Nor do I really get what problem that code would solve?

silasary: I ended up rewriting the code so that this object wasn't needed

Wolfhound905:
Any update Polls?

LordOfPolls:
I'd point out you never said this PR was now done, and it has merge conflicts.

Wolfhound905:
Can't you just click resolve? It seems like a very simple thing to fix.
https://github.com/Discord-Snake-Pit/Dis-Snek/pull/112/conflicts
image
also, I am not sure what I am supposed to do here. This just looks like GitHub messin up

LordOfPolls:
I could, but like I said; you never told me this PR was done so I've done nothing since your message 4 days ago.
It's ten at night, I'll look at this tomorrow.

i0bs:
Btw resolve is for the branch to remove merge conflicts to the actual PR commits, that would mess up the branch code

Wolfhound905:
I am not sure why the conflict is even there.

LordOfPolls:
Ideally the PR author should resolve conflicts. But the repo owner can resolve it without breaking the branch code.

The reason for the conflict is because both master and dev have been updated since you created your fork. And you've modified the same file that has been modified on master/dev

LordOfPolls (approved):

PR #118, opened 2021-12-19
Add search_guild_members

Kamisoi:

- [ ] Breaking code change

## Description

Added method for searching member of guild via string, also possible to get more than one by providing limit integer.

## Changes

Add async method for [search_guild_members](https://discord.com/developers/docs/resources/guild#search-guild-members)

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/guild.py
+++ b/dis_snek/models/discord_objects/guild.py
@@ -1017,6 +1017,10 @@ class Guild(BaseGuild):
         data = await self._client.http.get_guild_integrations(self.id)
         return [GuildIntegration.from_dict(d | {"guild_id": self.id}, self._client) for d in data]
 
+    async def search_guild_members(self, name: str, limit: int = 1) -> List["Member"]:
+        data = await self._client.http.search_guild_members(guild_id=self.id, query=name, limit=limit)
+        return [self._client.cache.place_member_data(self.id, _d) for _d in data]
+
 
 @define()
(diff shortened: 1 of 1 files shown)


dis_snek/models/discord_objects/guild.py
```diff
@@ -1017,6 +1017,10 @@ async def get_guild_integrations(self) -> List["GuildIntegration"]:
         data = await self._client.http.get_guild_integrations(self.id)
         return [GuildIntegration.from_dict(d | {"guild_id": self.id}, self._client) for d in data]
 
+    async def search_guild_members(self, name: str, limit: int = 1) -> List["Member"]:
```
Kamisoi: After some thought here it shouldn't be "name" but something else, since it can be any string that could be used to find user nickname or username.

artem30801:
IMO it should be named just `search_members`, given that it's a Guild method

dis_snek/models/discord_objects/guild.py
LordOfPolls: Query?

Kamisoi: How about search_query, for even more clarification? (at least imo)

LordOfPolls: Seems kinda redundant seeing as Query means to ask for information, and the method is literally called search

PR #129, opened 2021-12-27
feat: Redesign the gateway to be safe against race conditions

Bluenix2:
## Description

Previously there was all sorts of issues with the gateway and band-aid solutions for them all..

The purpose of this PR is to remove all of those for a system that should work much more safely. It adds more locks and events that make the code a bit more complicated, but it should be easier to follow.

## Changes

- Move the `BeeGees` class into a task on the class
- Make the `WebsocketClient` class an asynchronous context manager
- Update the scope and usage of `WebsocketClient` such that it now reconnects automatically
- Guard operations with a lock and event automatically
- Ensure `WebsocketClient` is safe against race conditions and handles cancellation with the asynchronous context manager

## Additional notes

This is opened as a draft as it is not yet tested (or that the race conditions that are supposed to have been solved, in fact are solved).

### DIFF
--- a/dis_snek/gateway.py
+++ b/dis_snek/gateway.py
@@ -8,9 +8,10 @@ import random
 import sys
 import time
 import zlib
+from contextlib import asynccontextmanager
 from typing import Any, List, Optional, TYPE_CHECKING
 
-from aiohttp import WSMsgType
+from aiohttp import WSCloseCode, WSMsgType, ClientWebSocketResponse
 
 from dis_snek.const import logger_name, MISSING
 from dis_snek.errors import WebSocketClosed
@@ -143,9 +144,9 @@ class WebsocketClient:
     #     "_chunk_lock",
     # )
 
-    def __init__(self, session_id: Optional[int] = None, sequence: Optional[int] = None) -> None:
-        self.session_id = session_id
-        self.sequence = sequence
+    def __init__(self, state: "ConnectionState", ws: ClientWebSocketResponse) -> None:
+        self.state = state
+        self.ws = ws
 
         self.buffer = bytearray()
         self._zlib = zlib.decompressobj()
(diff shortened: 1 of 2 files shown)


dis_snek/gateway.py
```diff
@@ -39,177 +42,118 @@ async def rate_limit(self) -> None:
...
+        self._entered = True
+
+        self.ws = await self.state.client.http.websocket_connect(self.state.gateway_url)
+        self._closed.set()
+
+        hello = await self.receive()
+        self.heartbeat_interval = hello["d"]["heartbeat_interval"] / 1000
+
+        await asyncio.sleep(self.heartbeat_interval * random.uniform(0, 0.5))
+        self._keep_alive = asyncio.create_task(self._start_bee_gees())
```
LordOfPolls: This delays both heartbeating and indentify. Per api spec, only the heartbeat needs delaying by `heartbeat_interval * jitter`.

LordOfPolls (changes requested):
Looks really good, i only had one gripe with it.
I have also tested it, and it seems to work perfectly

dis_snek/gateway.py
Bluenix2: Yes I originally read it like this but whilst implementing it here I realized that surely Discord meant both the heartbeating and IDENTIFY. Heartbeating itself can't be that costly for Discord - IDENTIFY is on the other hand.

[The documentation here](https://discord.com/developers/docs/topics/gateway#identifying) implies that after starting to heartbeat (which happens after that original delay) you should IDENTIFY - hence the wait delays both the heartbeat and IDENTIFY.

I am not sure if I am right in this, but it's not much to change either way.

LordOfPolls: I don't believe your interpretation is correct. You send your receive hello, then heartbeat_interval * jitter later you send heartbeats. I don't believe identify is included in that delay.

LordOfPolls (approved):

PR #133, opened 2021-12-30
fix: Resolve some bad variable references, and add linter to ensure they can't happen again

silasary:

- [ ] Breaking code change

## Description
In the course of 76f6d05f32fb9e445f49895cb4613368e341637a, Polls accidentally partially reintroduced the old name for `rate_limit_per_user`
I've fixed that, and added a bare-bones flake8 config and linter to ensure errors like that can't return.

## Checklist

- [x] I've tested my code
- [x] This PR can be rebased

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,9 +3,16 @@
 # top-most EditorConfig file
 root = true
 
-[*.py]
+[*.(py|md)]
 indent_style = space
 indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.yml]
+indent_style = space
+indent_size = 2
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,12 @@
+[flake8]
+extend-ignore = 
+    Q0, E501, C812, E203, W503  # These default to arguing with Black.  We might configure some of them eventually
+    F403, # F403 tells us not to use * imports.  It's correct, but it'll be work to fix.
+    ANN1, # These insist that we have Type Annotations for self and cls.
+    ANN204, ANN206, # return annotations for special methods and class methods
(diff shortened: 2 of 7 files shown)


AstreaTSS:
I want to see what others think personally, but I think this is a bit *too* much for one PR.

Obviously, the bad variable references should be fixed, but adding a linter should require another PR, at least to me. It also touches `.editorconfig` for what seems like no reason.

Also, there's already a pre-commit hook for flake. I think that would have caught it, honestly, but I think the problem is that every time it's used, people ignore it because the library is *definitely* not close to how flake8 wants things to be. Either we should address every issue the hook brings up, or adjust the hook itself to be more minimal so that it still can have a use.

...though, that *was* kinda a rant and it's probably best to discuss that in another PR or in the Discord server.

silasary:
> Also, there's already a pre-commit hook for flake. 

I don't know how many people use pre-commit. Having it run as an action is definitely were we want it to be.

> I think that would have caught it, honestly, but I think the problem is that every time it's used, people ignore it because the library is definitely not close to how flake8 wants things to be. 

Hence why I tweaked the settings to turn off every warning we currently violate.

AstreaTSS:
Yeah, that's fair enough now since I take a deeper look at it. I still have slight concerns about this PR potentially doing too much either way (though that may just be me), but I *definitely* support making it an action.

Maybe the pre-commit hook could also be adjusted to account to turn off settings we violate? I feel like that would help out with all of this.

silasary:
> Maybe the pre-commit hook could also be adjusted to account to turn off settings we violate? I feel like that would help out with all of this.

No explicit adjustments are needed.  It'll pull the same settings out of the `.flake8` file

AstreaTSS:
...my brain is not working today 😆.

Though the arguments for the hook should probably still be adjusted to remove the currently ignored settings, right? I mean, the config will take care of that anyways.

silasary:
Good catch, I didn't realize we were explicitly setting an ignore there.

PR #143, opened 2022-01-04
Feat: Add helper method to return a guild member

Kigstn:

- [ ] Breaking code change

## Description

 Add helper method to return a guild member

## Changes

- add `Guild.get_member()`

## Checklist

- [ ] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/guild.py
+++ b/dis_snek/models/discord_objects/guild.py
@@ -295,6 +295,18 @@ class Guild(BaseGuild):
                 return role
         return None
 
+    async def get_member(self, member_id: "Snowflake_Type") -> Optional["Member"]:
+        """
+        Return the Member with the given discord ID
+
+        Args:
+            member_id: The ID of the member:
+        Returns:
+            Member object or None
+        """
+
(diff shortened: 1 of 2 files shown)


silasary (approved):

dis_snek/models/discord_objects/guild.py
```diff
@@ -295,6 +295,18 @@ def my_role(self) -> Optional["Role"]:
...
+        """
+        Return the Member with the given discord ID
+
+        Args:
+            member_id: The ID of the member:
+        Returns:
+            Member object or None
+        """
+
+        return await self._client.get_member(user_id=member_id, guild_id=self.id)
```
artem30801: IMO this should be consistent with `get_owner` method (as in, `return await self._client.cache.get_member(self.id, member_id)` )

artem30801 (changes requested):

dis_snek/models/discord_objects/guild.py
Kigstn: Sure, makes sense

silasary (approved):

artem30801 (approved):

PR #147, opened 2022-01-05
docs: Annotated even more return types

silasary:

- [ ] Breaking code change

## Description
Progress towards ANN201 (Missing return type annotation for public function)

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/debug_scale/__init__.py
+++ b/dis_snek/debug_scale/__init__.py
@@ -36,7 +36,7 @@ class DebugScale(DebugExec, DebugAppCMD, DebugScales, Scale):
         sub_cmd_name="info",
         sub_cmd_description="Get basic information about the bot",
     )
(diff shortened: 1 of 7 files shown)


artem30801:
Should it be None or typing.NoReturn though? I think NoReturn was used in several places

LordOfPolls:
NoReturn means the function will never terminate, and if it does it will raise an exception.

It should not be used by functions that lack a return line, they would return None

I believe NoReturn has been used erroneously in some recent commits

silasary:
I haven't used it here, but yeah, I can do an audit of where we have used it

silasary:
Yeah, we only use NoReturn in one place: `Snake._raise_sync_exception`

LordOfPolls:
All good then

PR #151, opened 2022-01-08
fix: Return None if the cache.get_channel's fallback request errors

silasary:

- [ ] Breaking code change

## Description
`cache.get_channel` was already marked as Optional, and if called with `request_fallback=False`, would return None instead of throwing an error.
This PR adds consistency when called with `request_fallback=True` on a channel that is hidden from the bot. 
Most notably, this fixes an error that would happen when generating an InteractionContext in a channel/guild without the bot user.

## Checklist

- [x] I've tested my code


LordOfPolls:
Imo, it should raise if you've asked it to request from the API and it hasn't found it

PR #157, opened 2022-01-10
feat: Defer the unpacking of guild channel and members

silasary:

- [ ] Breaking code change

## Description
Makes Guild instantiation much lighter by throwing Channels and Members into the cache as deferred dicts (that get inflated when first pulled from the cache)

This is non-breaking, with the exception of people bypassing GlobalCache's methods to access the underlying dicts directly

Things I thought about doing, but have not done:
- Background task that slowly inflates deferred stuff while it's idle in the cache.  Discarded because it didn't feel needed.
- Deferring Roles.  The speedup from Members and Channels alone was good enough in my benchmarks.

## Changes
- Members and Channels can be optionally cached as Deferred dicts, rather than being constructed upon storage.
- The Guild constructor uses this to speed up initial load times
- The cache objects for `member_cache` and `channel_cache` are now private.
- Synchronous functions to access these caches have been added (Though I'm not 100% happy with the names)

## Checklist

- [x] I've tested my code


silasary:
Benchmarks tell me this approach is a waste of time, restarting from a different idea

PR #163, opened 2022-01-12
docs(enums): Correct grammatical errors.

i0bs:

- [ ] Breaking code change
- [X] Documentation change/addition

## Description
I looked at some of the documentation for your enumerable objects and found that you had grammatical inconsistencies in the word usage. This PR aims to correct that.

## Changes
* Used proper English grammar.

## Checklist

- [X] I've formatted my code with [Black](https://black.readthedocs.io/en/stable/)
- [X] I've ensured my code works on `Python 3.10.x`
- [X] I've tested my code

(I didn't test it at all but let's go with that 😉)

### DIFF
--- a/dis_snek/models/enums.py
+++ b/dis_snek/models/enums.py
@@ -534,7 +534,7 @@ class ActivityType(IntEnum):
 
 
 class Status(str, Enum):
-    """Represents the status' a user may have."""
+    """Represents the statuses a user may have."""
 
     ONLINE = "online"
     OFFLINE = "offline"
--- a/dis_snek/models/enums.py
+++ b/dis_snek/models/enums.py
@@ -160,7 +160,7 @@ class UserFlags(DiscordIntFlag):  # type: ignore
     HOUSE_BALANCE = 1 << 8
     """User belongs to the `balance` house"""
     EARLY_SUPPORTER = 1 << 9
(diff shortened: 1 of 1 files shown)


dis_snek/models/enums.py
```diff
@@ -534,7 +534,7 @@ class ActivityType(IntEnum):
 
 
 class Status(str, Enum):
-    """Represents the statuses a user may have."""
+    """Represents the status' a user may have."""
```
silasary: Errant apostrophe

silasary (changes requested):

dis_snek/models/enums.py
i0bs: Disagree, it's valid in punctuation and in it's contextually derived use it's proper for plural representation of how many status is possessed.

LordOfPolls: In English statuses is acceptable as per the Oxford Dictionary

i0bs: Well I think Oxford is wrong >:(

LordOfPolls: This PR wont be merged until this is reverted

i0bs: ok ok fine :(

silasary (approved):

PR #164, opened 2022-01-12
feat: Improve Role comparison

leestarb:
"__le__" and "__ge__" are also added

- [ ] Breaking code change

## Description

Enhance behavior of Role comparison

## Changes

- [x] Validation checks and  type hints

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/role.py
+++ b/dis_snek/models/discord_objects/role.py
@@ -54,12 +54,6 @@ class Role(DiscordObject):
     def __gt__(self: "Role", other: "Role") -> bool:
         return Role.__lt__(other, self)
 
-    def __le__(self: "Role", other: "Role") -> bool:
-        return not Role.__lt__(other, self)
-
-    def __ge__(self: "Role", other: "Role") -> bool:
-        return not Role.__lt__(self, other)
-
     @classmethod
     def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
         data.update(data.pop("tags", {}))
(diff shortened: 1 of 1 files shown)


artem30801:
Very nice, though initially, I left out le and ge, because we have eq reserved for id comparison - implementation of comparisons by different properties may lead to serious confusion of the users.
I want to hear the opinion of other contributors/users about this

leestarb:
> Very nice, though initially, I left out le and ge, because we have eq reserved for id comparison - implementation of comparisons by different properties may lead to serious confusion of the users.
> I want to hear the opinion of other contributors/users about this

Yes, however with le and ge roles position comparison becomes more native. I think, it's nice if position check hasn't less precision than by id.

AlbertUnruh:
If my understanding is right, it doesn't matter if we check for `id` or `position` in \_\_eq\_\_, since every `position` should only be assigned once and if `A.position == B.guild` then should `A.id == B.id` also be `True`.

silasary:
Yeah, we should include le and ge

artem30801 (approved):
LGTM

artem30801 (changes requested):
Upon further review and investigation of python documentation:
> There are no other implied relationships among the comparison operators or default implementations; for example, the truth of (x<y or x==y) does not imply x<=y.


I would recommend to remove `gt` dunder and decorate the whole class with `@functools.total_ordering`

See:
https://docs.python.org/3/library/functools.html#functools.total_ordering

https://docs.python.org/3/reference/datamodel.html#object.__ge__

leestarb:
Maybe remove unnecessary Role type hints?

silasary:
There's no harm in having them

silasary (approved):

artem30801 (approved):

PR #169, opened 2022-01-14
Embed.__len__ optimizations

Vioshim:

- [ ] Breaking code change

## Description
Taking into account that embeds are constantly using the `__len__` property for validators whenever they are modified, the implementation of maps and optimal methods can help to make the validation faster which benefits with a gain 20-25% in comparison to the usual approach.

### Benchmark (tested with timeit, `n=1000000`)

| Embed | Default `__len__` | Modified `__len__` |
| - | - | - |
| Empty Embed | 0.2745 s | 0.1712 s |
| All parameters | 4.8217 s | 4.2479 s |
| All parameters (No Fields) | 0.7127 s | 0.6124 s

## Changes

• Modified `Embed.__len__`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord_objects/embed.py
+++ b/dis_snek/models/discord_objects/embed.py
@@ -209,16 +209,16 @@ class Embed(DictSerializationMixin):
         # yes i know there are far more optimal ways to write this
         # its written like this for readability
         total: int = 0
-        if title := self.title:
-            total += len(title)
-        if description := self.description:
-            total += len(description)
-        if footer := self.footer:
-            total += len(footer)
-        if author := self.author:
-            total += len(author)
-        if fields := self.fields:
-            total += sum(map(len, fields))
+        if self.title:
+            total += len(self.title)
+        if self.description:
+            total += len(self.description)
+        if self.footer:
+            total += len(self.footer)
+        if self.author:
+            total += len(self.author)
+        if self.fields:
+            total += sum(map(len, self.fields))
         return total
 
     def set_author(
--- a/dis_snek/models/discord_objects/embed.py
+++ b/dis_snek/models/discord_objects/embed.py
@@ -208,14 +208,17 @@ class Embed(DictSerializationMixin):
     def __len__(self):
         # yes i know there are far more optimal ways to write this
         # its written like this for readability
-        total = 0
-        total += len(self.title) if self.title else 0
-        total += len(self.description) if self.description else 0
-        total += len(self.footer) if self.footer else 0
-        total += len(self.author) if self.author else 0
-
-        for _field in self.fields:
-            total += len(_field)
+        total: int = 0
+        if title := self.title:
+            total += len(title)
+        if description := self.description:
+            total += len(description)
+        if footer := self.footer:
+            total += len(footer)
+        if author := self.author:
+            total += len(author)
+        if fields := self.fields:
+            total += sum(map(len, fields))
         return total
 
     def set_author(


LordOfPolls:
You're attempting to merge to master, are you sure thats what you want to do?

Vioshim:
Just realized the mistake oops, was gonna use the dev branch

LordOfPolls (approved):

artem30801:
Looks nice
Does the walrus operator boost performance? Have you measured it in comparison to bare ifs and attribute re-usage?
Also, you should probably delete that in-code comment left for the older implementation X)

Bluenix2:
After generating the length, why not save it to an attribute and whenever the embed is modified set it to None?

Then when `__len__` is called check for this attribute. It'll make the `__len__` lookup instant if nothing changed.

LordOfPolls:
@artem30801 
> you should probably delete that in-code comment left for the older implementation X)

arguably i think it still applies 😜

LordOfPolls:
@Bluenix2 
> After generating the length, why not save it to an attribute and whenever the embed is modified set it to None?

It was my understanding *most* python objects cache their len

Vioshim:
> Looks nice Does the walrus operator boost performance? Have you measured it in comparison to bare ifs and attribute re-usage? Also, you should probably delete that in-code comment left for the older implementation X)

I've been making multiple tests and it always gave better results to use the walrus operator instead of the if, I suppose it's by the fact that the If has to first execute the "function" or value of both sides and then assign it 

aka `a if condition else b` while using the walrus allows to only use the len function if it's needed, otherwise the function is never called which allows that small boost in speed.

artem30801:
> > Looks nice Does the walrus operator boost performance? Have you measured it in comparison to bare ifs and attribute re-usage? Also, you should probably delete that in-code comment left for the older implementation X)
> 
> I've been making multiple tests and it always gave better results to use the walrus operator instead of the if, I suppose it's by the fact that the If has to first execute the "function" or value of both sides and then assign it
> 
> aka `a if condition else b` while using the walrus allows to only use the len function if it's needed, otherwise the function is never called which allows that small boost in speed.

No, I meant like
```python
if self.title:
    total += len(self.title)
```
Would that perform better? Worse? Same?
Could you please test it?

Vioshim:
> Would that perform better? Worse? Same? Could you please test it?

@artem30801 I've tested and seems to perform actually faster, most likely due to not having to use extra variables.

I'll proceed to update the PR

Bluenix2:
> > After generating the length, why not save it to an attribute and whenever the embed is modified set it to None?
> 
> It was my understanding _most_ python objects cache their len

Yes, they do (the strings, lists and other built-in structures) but the whole body of `__len__` for Embed will still execute:
```python
>>> class AlwaysZero:
...     def __len__(self) -> int:
...             print('Accessed')
...             return 0
...
>>> zero = AlwaysZero()
>>> for _ in range(10):
...     len(zero)
...
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
Accessed
0
```

It is up to us to cache the result.

LordOfPolls:
> Yes, they do (the strings, lists and other built-in structures) but the whole body of `__len__` for Embed will still execute:

Oh i see what you mean, i thought you meant caching each len, my bad for not reading correctly

LordOfPolls (approved):

artem30801 (approved):
LGTM

PR #173, opened 2022-01-15
chore: Remove unnecessary import

leestarb:

- [ ] Breaking code change

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/snek/wait.py
+++ b/dis_snek/models/snek/wait.py
@@ -1,9 +1,11 @@
-from asyncio import Future
-from typing import Coroutine, Callable, Optional
+from typing import TYPE_CHECKING, Callable, Optional
+
+if TYPE_CHECKING:
+    from asyncio import Future
 
 
 class Wait:
-    def __init__(self, event: str, checks: Optional[Callable[..., bool]], future: Future):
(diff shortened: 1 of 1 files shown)


benwoo1110:
I think this is not needed anymore after the restructure and import changes?

leestarb:
> I think this is not needed anymore after the restructure and import changes?

The restructure did not affect this, changes are still actual.

LordOfPolls:
This seems completely redundant. TYPE_CHECKING does not need to be used here

leestarb:
> This seems completely redundant. TYPE_CHECKING does not need to be used here 

Hmm, why?

silasary:
TYPE_CHECKING is used to prevent circular imports.  
There's no performance or readability gains from using it, and its use can sometimes make refactoring harder.  

We want to avoid it where possible, and this is not a place where it needs to be used.

leestarb:
> TYPE_CHECKING is used to prevent circular imports. There's no performance or readability gains from using it, and its use can sometimes make refactoring harder.
> 
> We want to avoid it where possible, and this is not a place where it needs to be used.

Okay, I got you

PR #175, opened 2022-01-16
refactor💥: Complete restructure of library

LordOfPolls:

- [x] Breaking code change

## Description

This PR completely restrucutres the library into a layout that makes more sense, and was agreed to by the contributors. 
This PR also moves some methods and objects between classes and redoes the imports to help prevent circular imports during typing.

## Changes

- Restrucutre library into this format
```
  ├───api
  │   ├───events
  │   │   └───processors
  │   ├───gateway
  │   └───http
  │       └───http_requests
  ├───client
  │   ├───mixins
  │   └───utils
  ├───ext
  │   ├───debug_scale
  │   └───tasks
  └───models
      ├───discord
      └───snek
          └───annotations
```
- Move member chunk processing into guild object
- Remove `sentinel.py` and move its contents into `const.py`

## Checklist

- [x] I've tested my code


LordOfPolls:
The new docs structure -> https://dis-snek--175.org.readthedocs.build/
Reviews appreciated

AstreaTSS:
I misclicked, oops 😅

AstreaTSS:
Anyways, since my tablet is struggling to view every single file: is it possible to move the `Models` section above the `Ext` one in the docs? I just think that'll make more sense.

Otherwise, the docs look great!

LordOfPolls:
> is it possible to move the `Models` section above the `Ext` one in the docs? I just think that'll make more sense.

without specifying the entire pages structure, i dont believe i can, unless i slap numbers before each folder

AstreaTSS:
If you're not opposed to using a plugin, I believe the [awesome-pages plugin](https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin) can do it.

LordOfPolls:
Good find, added and done

Kigstn:
Looking through the docs, I have a few smaller points:

- I feel like the events `GuildEvent` /  `BaseEvent` should be removed from the docs, since they are solely inherited and cannot be used directly
- Links in the guides are broken now

Also, whats the status on having automatic links to models (f.e. When a method returns `Guild`, you should be able to click on that to get to the definition)

LGTM otherwise

LordOfPolls:
> - I feel like the events `GuildEvent` /  `BaseEvent` should be removed from the docs, since they are solely inherited and cannot be used directly

It's actually harder not to show them, plus they ought to be documented for people making their own events. 

> - Links in the guides are broken now

Which ones? I thought I had updated them 


> Also, whats the status on having automatic links to models (f.e. When a method returns `Guild`, you should be able to click on that to get to the definition)

Not something I can add, still waiting for mkdocstrings to add that functionality

artem30801:
> Not something I can add, still waiting for mkdocstrings to add that functionality

Such links cannot be automatically generated now but you can make them using full-import-path links in docstrings

Kigstn:
> Which ones? I thought I had updated them

The last one on [this](https://dis-snek--175.org.readthedocs.build/Guides/1%20Getting%20Started/) page for example. Probly did not happen because of the refactor, but I just now noticed :D

LordOfPolls:
> Such links cannot be automatically generated now but you can make them using full-import-path links in docstrings
 
See #lib-discussion

LordOfPolls:
> The last one on [this](https://dis-snek--175.org.readthedocs.build/Guides/1%20Getting%20Started/) page

Just checked. This PR didn't break that, I'll fix it in this regardless

leestarb:
👍 Very nice

LordOfPolls:
This PR should be ready for a full review

~~may god have mercy on your soul~~

Kigstn (approved):
After a thorough inspection of every line change, LGTM.

Completely unrelated: test bot starts of fine with this and I cannot find any issues in the docs.

LordOfPolls:
Because its a massive PR, and I really don't expect a line by line review 

Your checklist to review would be, 
- look through guild.py to make sure you're happy with the new Imports which I've used to prevent circular imports (there are other files I've changed with this method, but let's reduce workload) 
- Check the docs, make sure you're happy 
- Make sure you can run a bot off of this

dis_snek/models/discord/guild.py
```diff
@@ -209,57 +195,57 @@ def _process_dict(cls, data, client):
         return data
 
     @property
-    def channels(self) -> List["TYPE_GUILD_CHANNEL"]:
+    def channels(self) -> List["models.TYPE_GUILD_CHANNEL"]:
```
benwoo1110, line 198: This is already a `typing.Union`, so I think can don't put within quotes just like what you did with `Snowflake_Type`?

LordOfPolls: That would make it try and import that file during initialisation, causing circular imports

benwoo1110: actually, what's the condition for putting string typing? (apart from when its a class) I think I still see a few instance of it being inconsistent

benwoo1110: > That would make it try and import that file during initialisation, causing circular imports

oh rightttt, those lovely stuff

LordOfPolls: It's not a string when there is a reasonable chance that it *wont* cause circular imports. IE you bring up `Snowflake_Type`, which now doesnt depend on any other major model, meaning it can be outside of a string, and imported directly

PR #179, opened 2022-01-17
feat: Added more imports for ease-of-use

AstreaTSS:

- [ ] Breaking code change

## Description
With the reorganization, some files that were previously very simple to import now required quite lengthy code to do so. This PR fixes that by adding more imports in.

## Changes
- Added importing `events` (via `dis_snek.events`, like how it was before the restructuring) and `ext` (via `dis_snek.ext`).

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/__init__.py
+++ b/dis_snek/__init__.py
@@ -1,2 +1,4 @@
 from .client import *
 from .models import *
+from .api import *
+from . import ext
--- a/dis_snek/api/__init__.py
+++ b/dis_snek/api/__init__.py
@@ -1,0 +1,1 @@
+from . import events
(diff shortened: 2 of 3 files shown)


AstreaTSS:
Seems like something about this broke the RTD test. I'm busy right now - I'll look into it in a bit!

LordOfPolls:
Its broken because youve added `ext` to the root namespace, which it was never intended to do

AstreaTSS:
> Its broken because youve added `ext` to the root namespace, which it was never intended to do

lame smh

But yeah, I can see that. Still busy right now, but it *should* be a one-line fix, thankfully.

AstreaTSS:
There we go! Seems like importing `paginators` from `ext` early was causing the problem, so I fixed that quickly.
(...by not importing it. Shhhh.)

PR #187, opened 2022-01-17
docs: Update the component usage guide to include another callback method

Kigstn:

- [ ] Breaking code change

## Description

Update the component usage guide to include another callback method

## Changes

- add guide for `@component_callback(custom_id)`

## Checklist

- [ ] I've tested my code

### DIFF
--- a/docs/src/Guides/5 Components.md
+++ b/docs/src/Guides/5 Components.md
@@ -55,7 +55,7 @@ If you want to define the layout yourself, you have to put them in an `ActionRow
     await channel.send("Look, Buttons!", components=components)
     ```
 
-For simplicity's sake, example three will be used for all examples going forward. 
+For simplicity's sake, example three will be used for all examples going forward.
 
 If you want to delete components, you need to pass `components=[]` to `.edit()`.
 
--- a/docs/src/Guides/5 Components.md
+++ b/docs/src/Guides/5 Components.md
@@ -205,7 +205,7 @@ When responding to a component you need to satisfy discord either by responding 
     If you have a lot of components, putting everything in the `on_component()` event can get messy really quickly.
 
     Similarly to Option 1, you can define `@component_callback` listeners. This works after restarts too.
-    
+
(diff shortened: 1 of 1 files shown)


docs/src/Guides/5 Components.md
```diff
@@ -201,10 +201,37 @@ When responding to a component you need to satisfy discord either by responding
...
+                    style=ButtonStyles.GREEN,
+                    label="Click Me",
+                    disabled=False,
+                )
+            )
+        ]
+
+        await channel.send("Look a Button!", components=components)
+
+    @component_callback("my_button")
```
silasary: I'd put comment here mentioning that the parameter is the parameter is the custom ID

benwoo1110: maybe to be clearer, also change `"my_button"` to `"my_button_id"`

benwoo1110 (approved):
looks good

artem30801:
Looks good, but for simplicity, I'd advise pointing out that you can actually pass just one component or actionrow to `components` kwarg. Would also make code a LOT less scary in all examples

Kigstn:
> Looks good, but for simplicity, I'd advise pointing out that you can actually pass just one component or actionrow to `components` kwarg. Would also make code a LOT less scary in all examples

Fair, I'll update that

artem30801:
Nice, thank you
One more question - why do all the examples use `disabled=False` in Button creation? I don't think it's necessary in most cases

Kigstn:
> Nice, thank you One more question - why do all the examples use `disabled=False` in Button creation? I don't think it's necessary in most cases

Good point!

artem30801 (approved):
LGTM

silasary (approved):
Yeah, I'm happy with this.

But if I were to nitpick, I'd point out that Persistent Callback Option 3 should use an Action Row, because the complexity of that option is only valuable when you have multiple buttons

Kigstn:
> But if I were to nitpick, I'd point out that Persistent Callback Option 3 should use an Action Row, because the complexity of that option is only valuable when you have multiple buttons

While I agree, I would keep it this way for consistency sake

PR #192, opened 2022-01-18
Yet more flake8 progress

silasary:

- [ ] Breaking code change

## Changes
- style: C816 missing trailing comma in Python 3.6+
- style: R503 missing explicit return at the  end of function ableto return non-None value.
- docs: Yet more return type annotations

## Checklist

- [ ] I've tested my code

### DIFF
--- a/dis_snek/models/discord/role.py
+++ b/dis_snek/models/discord/role.py
@@ -1,5 +1,5 @@
 from functools import partial, total_ordering
-from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional, Union, TypeVar
 
 import attr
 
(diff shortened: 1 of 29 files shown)


dis_snek/models/discord/role.py
```diff
@@ -16,8 +16,10 @@
     from dis_snek.models.discord.user import Member
     from dis_snek.models.discord.snowflake import Snowflake_Type
 
+T = TypeVar("T")
 
-def sentinel_converter(value, sentinel=attr.NOTHING):
+
+def sentinel_converter(value: bool | T, sentinel: T = attr.NOTHING) -> bool:
```
artem30801: I think `value` should be also Optional as it can be None as per code

benwoo1110: Looking at how attr converter works, a value should always be passed into a converter method. (anyways also not really the point of this pr?)

artem30801: Optional == can be something OR `None`
And well, if this PR type hints this function, it better be more accurate?

benwoo1110: yea i suppose so

silasary: Good catch

PR #214, opened 2022-01-23
feat💥: Improve guild perms (see desc.)

leestarb:

- [x] Breaking code change

## Description

Improvements for the permissions related guild stuff

## Changes

- Fixed `Guild.permissions`
- Rewrite `member.guild_permissions` as a property
- Allow pass Snowflake_Type in `guild.is_owner` method 
- Documentate `guild.is_owner`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -312,8 +312,17 @@ class Guild(BaseGuild):
         # maybe precache owner instead of using `get_owner`
         return await self._client.cache.get_member(self.id, self._owner_id)
 
-    def is_owner(self, member: "models.Member") -> bool:
-        return self._owner_id == member.id
+    def is_owner(self, user_id: Snowflake_Type) -> bool:
+        """
+        Whether the user is owner of the guild.
+
+        Args:
+            user_id: The ID of the user
+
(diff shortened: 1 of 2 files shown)


dis_snek/models/discord/guild.py
```diff
@@ -312,8 +312,17 @@ async def get_owner(self) -> "models.Member":
         # maybe precache owner instead of using `get_owner`
         return await self._client.cache.get_member(self.id, self._owner_id)
 
-    def is_owner(self, member: "models.Member") -> bool:
-        return self._owner_id == member.id
+    def is_owner(self, user_id: Snowflake_Type) -> bool:
```
benwoo1110: You should Union `models.Member` here, also not sure about renaming the parameter to user_id, since the main idea is whether a member in that guild is the owner.

artem30801: As per definition: `Snowflake_Type = Union[int, str, "SnowflakeObject", SupportsInt]` you don't need to union Snowflake_Type with Member or anything, it's already included
But I agree about renaming back to `user`

artem30801 (approved):

PR #217, opened 2022-01-24
HTTP Request Improvements

LordOfPolls:

- [ ] Breaking code change

## Description

Improves http requests to be less smelly, and have more elegant control over ratelimit buckets. 

## Changes

- Adds more functionality to Route object
- Remove categorically ugly logic for bucket rate limiting
- Reduce attempts from 5 -> 3
- Ensure lock is always released for endpoints
- Fix incorrect endpoint being used to remove reactions (not necessary but why not)

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/api/http/http_client.py
+++ b/dis_snek/api/http/http_client.py
@@ -37,6 +37,8 @@ log = logging.getLogger(logger_name)
 
 
 class GlobalLock:
+    """Manages the global ratelimit"""
+
     def __init__(self) -> None:
         self.cooldown_system: CooldownSystem = CooldownSystem(
             45, 1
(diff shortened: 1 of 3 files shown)


dis_snek/api/http/http_client.py
```diff
@@ -49,6 +51,44 @@ async def rate_limit(self) -> None:
...
+
+    async def blind_deferred_unlock(self, delta: float) -> None:
+        """
+        Unlocks the BucketLock but doesn't wait for completion
+
+        args:
+            delta: The time until unlock, in seconds
+        """
+        self.unlock_on_exit = False
+        loop = asyncio.get_event_loop()
```
Bluenix2:
```suggestion
        loop = asyncio.get_running_loop()
```

dis_snek/api/http/http_client.py
```diff
@@ -118,98 +158,71 @@ async def request(
...
+                except OSError as e:
+                    if attempt < self._max_attempts - 1 and e.errno in (54, 10054):
+                        await asyncio.sleep(1 + attempt * 2)
+                        continue
+                    raise
+                except (Forbidden, NotFound, DiscordError, HTTPException):
+                    raise
+                except Exception as e:
+                    log.error("".join(traceback.format_exception(type(e), e, e.__traceback__)))
+                    break
```
Bluenix2: What's wrong with just letting it propogate?

dis_snek/api/http/route.py
```diff
@@ -1,27 +1,54 @@
 from typing import TYPE_CHECKING, Any, ClassVar, Optional
 from urllib.parse import quote as _uriquote
 
+
 if TYPE_CHECKING:
     from dis_snek.models.discord.snowflake import Snowflake_Type
 
 
 class Route:
```
Bluenix2, line 9: This should probably also implement `__eq__()`, I've made that mistake myself..

LordOfPolls: Why would you need to check equality of a Route?

dis_snek/api/http/http_client.py
LordOfPolls: Valid point, i believe this was left over from the previous smelly code. I'll remove this except

dis_snek/api/http/route.py
Bluenix2: You made it hashable, if you don't implement `__eq__()` then a Route can't be stored in a dictionary (well it can, but it falls back to `object()`'s implementation of `is` checks).

LordOfPolls: Very valid point, i'll add that

silasary (approved):

PR #220, opened 2022-01-25
fix💥: Set default channel position to last

leestarb:

- [x] Breaking code change

## Description

By default Discord creates channel at the bottom

## Changes

- Replace `0` with `None` (last)

## Checklist

- [x] I've tested my code


dis_snek/api/http/http_requests/channels.py
```diff
@@ -72,7 +72,7 @@ async def create_guild_channel(
         name: str,
         channel_type: Union["ChannelTypes", int],
         topic: Absent[Optional[str]] = MISSING,
-        position: int = 0,
+        position: Optional[int] = None,
```
silasary: Absent[int] = MISSING

silasary (changes requested):

dis_snek/api/http/http_requests/channels.py
leestarb: Are you sure?

artem30801: Why Absent tho? I think Optional in this context is fine

leestarb: Maybe Absent[Optional[int]] = MISSING?

LordOfPolls: This doesn't need to be missing, surely

LordOfPolls: We really need to nack down which to use `Absent` or `Optional` in context like this.

LordOfPolls (approved):
We need to come to a detrermination of where absent vs optional should be used, but that shouldnt block this pr. Im happy with it merging. Just waiting on @silasary

benwoo1110 (approved):

silasary (approved):

PR #227, opened 2022-01-27
fix💥: Broken guild get/fetch thread

leestarb:

- [x] Breaking code change

## Description

- Fix bug guild get/fetch thread methods return `None` always

## Changes

- Remove unnecessary `in` check
- Improve the docs
- 💥Channel methods no longer return `Thread`
- Allow `Snowflake_Type` everywhere

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -1000,7 +1000,7 @@ class Guild(BaseGuild):
             Channel object if found, otherwise None
 
         """
-        if channel_id in self._channel_ids and channel_id not in self._thread_ids:
+        if channel_id in self._channel_ids or channel_id in self._thread_ids:
             # theoretically, this could get any channel the client can see,
             # but to make it less confusing to new programmers,
             # i intentionally check that the guild contains the channel first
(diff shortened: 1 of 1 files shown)


dis_snek/models/discord/guild.py
```diff
@@ -1035,13 +1035,13 @@ def get_thread(self, thread_id: Snowflake_Type) -> Optional["models.TYPE_THREAD_
...
+
+        Note:
+            get_channel can retrieve threads, so this is basically an alias with extra steps for that
 
         """
-        # get_channel can retrieve threads, so this is basically an alias with extra steps for that
-        if thread_id in self._thread_ids:
-            return self.get_channel(thread_id)
-        return None
+        return self.get_channel(thread_id)
```
benwoo1110: maybe we should show an error if the channel is valid, but not a thread type channel? Bcu now is a bit confusing that this method will basically allow any type of channel to be fetched.

Kigstn: I suppose `NotFound` would be the most fitting

dis_snek/models/discord/guild.py
```diff
@@ -1000,16 +998,14 @@ def get_channel(
             Channel object if found, otherwise None
 
         """
-        if channel_id in self._channel_ids or channel_id in self._thread_ids:
+        if channel_id in self._channel_ids:
```
benwoo1110, line 1002: I think we should call the `to_snowflake` method to sanitise `channel_id`, bcu ppl may pass in as str instead of int, which may cause comparison issues.

leestarb: Sure.
Btw there are many times we missed this

LordOfPolls (approved):

benwoo1110 (approved):

PR #229, opened 2022-01-28
Dunder "all" the files

LordOfPolls:

- [x] Breaking code change

## Description

This PR adds dunder all to *most* of the library's files.

Breaking as it removes things from the root namespace that were never meant to be there in the first place. The biggest thing is that errors are no longer eroniously in the root namespace.

This PR will be squashed, dont worry.

## Changes

- Typed `__all__` too many times

## Checklist

- [x] I've tested my code


silasary (approved):
Heck yeah

leestarb:
Maybe use tuples?

LordOfPolls:
I dont think it should be. A list, normally, represents a sequence of the same type of item, a tuple typically represents a list of different types of item. 
For `__all__` it represents a list of object names, all being strings, so a list is most appropriate. 

edit: I'd point out this isnt enforced by python, purely idiomatic usage

PR #232, opened 2022-01-29
perf: Remove unnecessary cache check

leestarb:

- [ ] Breaking code change

## Description

I don't see that's necessary

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client/smart_cache.py
+++ b/dis_snek/client/smart_cache.py
@@ -284,8 +284,8 @@ class GlobalCache:
             if message.channel is None:
                 await self.get_channel(channel_id, request_fallback=True)
 
-            if not message.guild and isinstance(message.channel, GuildChannel) and message.channel.guild:
-                message._guild_id = message.channel.guild.id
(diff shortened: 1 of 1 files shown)


LordOfPolls:
If `message.channel.guild` is already not None, then nothing needs to be done, removing the check means we're unnecessarily always overwriting the ID which is more expensive than checking the value

LordOfPolls:
What would we do if the channel has no guild attribute set, and you try and access the guild. Removing this check opens up potential errors

leestarb:
> What would we do if the channel has no guild attribute set, and you try and access the guild. Removing this check opens up potential errors

I don't think that this check did something because:
- `message.guild` is `None`
- channel should be guilded
- if there's no `_guild_id` in the channel, this check won't work and nothing changed

LordOfPolls:
Resonable

silasary:
Yeah, I should have removed that check when I was adding the `isinstance` check.  Thanks for catching it

PR #237, opened 2022-01-30
docs: Sync type annotation

leestarb:

- [ ] Breaking code change

## Checklist

- [ ] I've tested my code

### DIFF
--- a/dis_snek/models/snek/listener.py
+++ b/dis_snek/models/snek/listener.py
@@ -49,5 +49,5 @@ class Listener:
         return wrapper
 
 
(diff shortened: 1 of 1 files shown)


silasary (approved):

silasary (changes requested):
I don't think this works.  You need to annotate that you're passing the type itself, not an object of said type

leestarb:
> I don't think this works.  You need to annotate that you're passing the type itself, not an object of said type

Not sure what you mean. That's just a copy of https://github.com/Discord-Snake-Pit/Dis-Snek/blob/dev/dis_snek/models/snek/listener.py#L27

PR #250, opened 2022-02-05
fix: behaviour of role add and delete events

artem30801:

- [X] Breaking code change

## Description
Reworked role add and delete events so they update all caches/related sets properly

## Changes

- role_create event now adds role id to guild._role_ids
- role_delete event now removes role id from global cache, guild._role_ids, member._role_ids
- RoleUpdate.before is now Absent["Role"] (fixed typehint)
- 💥RoleDelete.role_id renamed to RoleDelete.id
- Added role: Absent["Role"] to RoleDelete

## Checklist

- [X] I've formatted my code with [Black](https://black.readthedocs.io/en/stable/)
- [X] I've ensured my code works on `Python 3.10.x`
- [X] I've tested my code

### DIFF
--- a/dis_snek/api/events/processors/role_events.py
+++ b/dis_snek/api/events/processors/role_events.py
@@ -18,13 +18,18 @@ log = logging.getLogger(logger_name)
 class RoleEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_guild_role_create(self, event: "RawGatewayEvent") -> None:
-        g_id = event.data.get("guild_id")
-        role = self.cache.place_role_data(g_id, [event.data.get("role")])
-        self.dispatch(events.RoleCreate(g_id, role[int(event.data["role"]["id"])]))
+        g_id = int(event.data.get("guild_id"))
+        r_id = int(event.data["role"]["id"])
+
+        guild = self.cache.guild_cache.get(g_id)
+        guild._role_ids.add(r_id)
+
+        role = self.cache.place_role_data(g_id, [event.data.get("role")])[r_id]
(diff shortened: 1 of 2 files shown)


dis_snek/api/events/processors/role_events.py
```diff
@@ -35,4 +40,16 @@ async def _on_raw_guild_role_update(self, event: "RawGatewayEvent") -> None:
...
+        r_id = int(event.data.get("role_id"))
+
+        guild = self.cache.guild_cache.get(g_id)
+        role = self.cache.role_cache.pop(r_id, MISSING)
+
+        guild._role_ids.discard(r_id)
+
+        role_members = (member for member in guild.members if member.has_role(r_id))
+        for member in role_members:
+            member._role_ids.remove(r_id)
```
LordOfPolls, line 53: Imo this should be error-handled. If *somehow* the ID isnt already in the list, this will fail with a `ValueError`. Better to handle that edge case

artem30801: I don't quite agree because `member.has_role` checks if `role id` is in `member._role_ids`, and then it is removed from `member._role_ids` in for cycle. So there is no way for a real edge case here, as both segments of code check and assume the same thing. Given that there is nothing async involved, nothing sane modify data mid-cycle either

leestarb (approved):

dis_snek/api/events/processors/role_events.py
LordOfPolls: Reasonable point, resolving this thread

PR #256, opened 2022-02-08
FEAT: add property to voice channel that returns users currently in voice chat

Kigstn:

- [ ] Breaking code change

## Description

Before there was no way to tell who is currently in a voice channel. This is not perfect, since if the bot starts when people are already in a channel, it won't get caught. Since discord does not dispatch voice members anywhere however, this is the best we can do pretty sure

## Changes

- add .voice_members property to VoiceChannel
- add / remove members from there on VoiceStateUpdate

## Checklist

- [x] I've tested my code


silasary (approved):
Neat.  I like it

Kigstn:
Alright so with the most recent commit, it got a little more wonky. Basically, `VoiceState` now looks if its `member_id` is actually a part of its `.channel` obj, and if not copies the obj and adds it locally.

This enables `VoiceStateUpdate.before` to display the correct number of `voice_members` before the event

PR #266, opened 2022-02-09
feat: Allow using event typehint to define listener

leestarb:

- [ ] Breaking code change

## Description

ie:
```python
@listen()
async def receive_msg(event: events.MessageCreate): ...
```

## Changes

- Use typehint as event name if any

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/snek/listener.py
+++ b/dis_snek/models/snek/listener.py
@@ -32,11 +32,14 @@ class Listener:
             name = event_name
 
             if name is MISSING:
-                if not coro.__name__.startswith("_"):
-                    for typehint in coro.__annotations__.values():
-                        if inspect.isclass(typehint) and issubclass(typehint, BaseEvent):
-                            name = typehint.__name__
-                            break
+                for typehint in coro.__annotations__.values():
+                    if (
+                        inspect.isclass(typehint)
+                        and issubclass(typehint, BaseEvent)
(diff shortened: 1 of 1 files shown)


Kigstn (approved):
Neat. Added docs for this in #267

dis_snek/models/snek/listener.py
```diff
@@ -32,7 +32,14 @@ def wrapper(coro: Coroutine) -> "Listener":
             name = event_name
 
             if name is MISSING:
-                name = coro.__name__
+                if not coro.__name__.startswith("_"):
```
artem30801: Why exactly is this necessary?

leestarb: The reason for this check is clients internal `_on_websocket_ready` method, that accepts `RawGatewayEvent` (invalid event name)
Also is a good stub for future

LordOfPolls: This would prevent users from using private methods for listeners (with the type hint decloration), which is a very common use case. 

Imo the order of operations ought to be 
`check for passed parameter dictating event` -> `check coroutine name` -> `check argument type hints`

This pr, assuming im reading it correctly would reverse the latter 2
`check for passed parameter dictating event` -> `check argument type hints` -> `check coroutine name`
imo this is not desirable and the new functionality wouldn't even work for private methods.

LordOfPolls (approved):

PR #268, opened 2022-02-09
FIX: embed parsing from discord data 

Kigstn:

- [ ] Breaking code change

## Description

It did not create the objs correctly before and instead the dict data into an obj 

## Changes

- fixed parsing

## Checklist

- [x] I've tested my code


Kigstn:
My branch was borked, sec

PR #272, opened 2022-02-10
FEAT: add ability to get mutual guilds / member instances for users

Kigstn:

- [ ] Breaking code change

## Description

This adds the ability to get the member objs directly from User. This also fixes BaseUser.mutual_guilds which did not work

## Changes

- add User.member_instances
- fix BaseUser.mutual_guilds

## Checklist

- [x] I've tested my code


dis_snek/models/discord/user.py
```diff
@@ -115,6 +116,14 @@ def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
 
         return data
 
+    @property
+    def member_instances(self) -> List["Member"]:
+        """Returns the member object for all guilds both the bot and the user are in. Note: This will only be accurate if the guild members are cached internally"""
```
leestarb: Notes format seems to be inaccurate
I suggest you using this:
```
??? note
    This will only be accurate if the guild members are cached internally
```

LordOfPolls: Specifically `!!!` is for notes `???` is for collapsable notes

Kigstn: Sure, thats better. I used `!!!` since that seems to be used most in the lib

PR #280, opened 2022-02-13
refactor: Use `attr_utils.str_validator` in embed field value validation

leestarb:

- [ ] Breaking code change

## Description

`str_validator` Provides a warning message

## Changes

- Use `attr_utils.str_validator` in embed field value validation

## Checklist

- [x] I've tested my code


LordOfPolls:
Does this actually need a warning? Imo this is the same as sending a non-str object with `ctx.send` which doesnt need a warning.

The purpose of `str_validator`'s warning for components is that the values/custom_id's are actively referenced and it is not intuitive if the type changes transparently, i don't believe embeds fit that.

LordOfPolls:
Very curious about peoples views on this

PR #288, opened 2022-02-15
fix: Paginator always trying to create SelectMenu

Miner28:

- [ ] Breaking code change

## Description
Paginator was always trying to create SelectMenu which caused any Paginator that had over 25 pages to crash.

## Changes
fix: Removed `all` from call to create_components

## Checklist

- [x] I've tested my code


LordOfPolls:
While this resolves the bug short term, i believe it would break paginators in certain use cases. ie

```python
p = Paginator(...)
p.show_select_menu  = True
```

In the above code snippet, as the custom id's are cached during initialisation, they would never be added to the callback. So unfortunately this will need a more substantial fix that first expected

PR #289, opened 2022-02-16
refactor: Make all gets non-async, all fetches async

zevaryx:

- [x] Breaking code change

## Description

Creates naming convention parity between the cache and all Discord models. Currently, many functions with `get_*` are async, when in reality they should be called `fetch_*`. This PR renames all of these, as well as adding non-async `get_*` where applicable.

## Changes

- Fix all `get_*` functions in `dis_snek.models.discord`
- Fix all `get_*` functions in `dis_snek.client`
- Add async `fetch_*` and non-async `get_*` methods where applicable

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -316,9 +316,9 @@ class Guild(BaseGuild):
         """Alias for me.guild_permissions"""
         return self.me.guild_permissions
 
-    async def get_member(self, member_id: Snowflake_Type) -> Optional["models.Member"]:
+    async def fetch_member(self, member_id: Snowflake_Type) -> Optional["models.Member"]:
         """
-        Return the Member with the given discord ID.
+        Return the Member with the given discord ID, fetching from the API if necessary.
 
         Args:
             member_id: The ID of the member:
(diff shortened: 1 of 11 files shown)


dis_snek/models/discord/guild.py
```diff
@@ -327,10 +327,36 @@ async def get_member(self, member_id: Snowflake_Type) -> Optional["models.Member
...
+        return self._client.cache.get_member(self.id, member_id)
+
+    async def fetch_owner(self) -> "models.Member":
+        """
+        Return the Guild owner
+
+        Returns:
+            Member object or None
+        """
+        # TODO: maybe precache owner instead of using `fetch_owner`
```
LordOfPolls, line 348: The owner user object is cached during startup. 

This method should still exist, just in case the user is no longer in the cache, for some reason, might be worth looking into protecting that cached object, but thats an aside

LordOfPolls (approved):
Names look fine, untested by me, however

leestarb:
Hey, thats breaking

dis_snek/models/discord/thread.py
```diff
@@ -30,25 +30,45 @@ class ThreadMember(DiscordObject, SendMixin):
...
+
+        """
+        return self._client.cache.get_channel(self.id)
+
+    async def fetch_user(self) -> "User":
+        """
+        Fetch the user associated with this thread member.
+
+        Returns:
+        The user object
```
leestarb: Missing tabs cause incorrect docs

leestarb (changes requested):

dis_snek/models/discord/thread.py
zevaryx: Fixed in latest commit

leestarb (changes requested):
I dont think that `cache.get_*` can raise not found
Hence you should remove excepts in gets

upd: added in e09c7d6940f4d82433653fb3246c057c05e0bb4b

leestarb (changes requested):
You have to check whether given instance is in the guild in `guild.get_*`

upd: added in 9152154618de433013c1b70a671599b51a65cf32

dis_snek/models/discord/guild.py
```diff
@@ -1015,6 +1054,19 @@ async def get_role(self, role_id: Snowflake_Type) -> Optional["models.Role"]:
...
+        Get the specified role by ID.
+
+        Args:
+            role_id: The ID of the role to get
+
+        Returns:
+            A role object or None if the role is not found.
+
+        """
+        return self._client.cache.get_role(self.id, role_id)
```
artem30801: cache.get_role only needs role_id

silasary (approved):

leestarb (approved):
Congratulations

PR #306, opened 2022-02-20
feat💥: allow for iterables of str for msg prefixes

AstreaTSS:

- [x] Breaking code change
- [x] Documentation change/addition (technically)

## Description

As of right now, message commands can have dynamic prefixes via `get_prefix`, and that is very useful. However, the problem is that `get_prefix` only allows for *one* prefix to be returned when it is used. This results in something like this happening:

```python
async def get_prefix(msg: dis_snek.Message):
    prefixes = database_handler.get_prefixes(msg.guild.id) # returns a list of prefixes
    for prefix in prefixes:
        if msg.content.startswith(prefix):
            return prefix
```

...to which `dis-snek` would check *again* if the message's content started with that prefix. Not to mention it's just not convenient at all, especially when `discord.py` didn't require this (and I know we *aren't* `discord.py`, but still).

In an ideal world, you would be able to do something like:

```python
async def generate_prefixes(msg: dis_snek.Message):
    return database_handler.get_prefixes(msg.guild.id)
```

...and let `dis-snek` handle it from there, finding out which of the prefixes were used. That's what this PR does.
(I type a lot, don't I?)

## Changes

- 💥 Change `get_prefix` to `generate_prefixes` for clarity.
  - Technically, this could be made non-breaking by leveraging `kwargs` when initing `Snake` and have it check for `get_prefix` if `generate_prefixes` is at its default. This hasn't been done before in this library, however, and I figure we're, well, in alpha, so might as well.
  - Current `dis-snek` bots will have to rename `get_prefix` to `generate_prefixes` for their init for their `Snake`s, but won't have to change much else.
- Allow for `default_prefix` to be `Iterable[str]`, and for `generate_prefixes` to return `Iterable[str]` (they both can still use `str`). You can see the implementation yourself, but at the end of the day, it's just a for loop.
- Import `collections.abc.Iterable` into `client.py` so things are typehinted correctly.

## Checklist

- [x] I've tested my code

## Additional Notes
- Static, one-entry prefixes (aka something like "s!" passed into `default_prefix` and nothing else) may *seem* like they're being treated inefficiently, having to be added to a set of just itself and for-looped from there. However, testing revealed that this resulted in *less than a millisecond* of delay compared to just checking the prefix directly. It's not worth adding a special case to bypass the for loop.
- Ignore the weird commit history - just squash this, please! `sync_interactions` being changed to be default `True` messed things up since I was working on this for a little while.

### DIFF
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -1283,7 +1283,7 @@ class Snake(
                     if mention:
                         prefix = mention.group()
                     else:
-                        return
+                        continue
 
                 if message.content.startswith(prefix):
                     prefix_used = prefix
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -1273,7 +1273,7 @@ class Snake(
             if isinstance(prefixes, str):
                 # its easier to treat everything as if it may be an iterable
                 # rather than building a special case for strings
-                prefixes = {prefixes}
+                prefixes = (prefixes,)
 
             prefix_used = None
 
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -8,6 +8,7 @@ import sys
 import time
 import traceback
 from datetime import datetime
+from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, Callable, Coroutine, Dict, List, NoReturn, Optional, Type, Union
 
 import dis_snek.api.events as events
@@ -140,7 +141,7 @@ class Snake(
         self,
         intents: Union[int, Intents] = Intents.DEFAULT,
         loop: Optional[asyncio.AbstractEventLoop] = None,
-        default_prefix: str = MENTION_PREFIX,
+        default_prefix: str | Iterable[str] = MENTION_PREFIX,
         generate_prefixes: Absent[Callable[..., Coroutine]] = MISSING,
         sync_interactions: bool = False,
         delete_unused_application_cmds: bool = False,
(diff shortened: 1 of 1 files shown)


AstreaTSS:
Ignore that little draft mess-up there - I mis-clicked 😅

silasary (approved):
Nice

dis_snek/client/client.py
```diff
@@ -1267,22 +1268,35 @@ async def _dispatch_msg_commands(self, event: MessageCreate) -> None:
...
-                invoked_name = get_first_word(message.content.removeprefix(prefix))
+            prefix_used = None
+
+            for prefix in prefixes:
+                if prefix == MENTION_PREFIX:
+                    mention = self._mention_reg.search(message.content)
+                    if mention:
+                        prefix = mention.group()
+                    else:
+                        return
```
benwoo1110: why return here? I think it should be `continue`

dis_snek/client/client.py
```diff
@@ -1267,22 +1268,35 @@ async def _dispatch_msg_commands(self, event: MessageCreate) -> None:
...
-            if prefix == MENTION_PREFIX:
-                mention = self._mention_reg.search(message.content)
-                if mention:
-                    prefix = mention.group()
-                else:
-                    return
+            if isinstance(prefixes, str):
+                # its easier to treat everything as if it may be an iterable
+                # rather than building a special case for strings
+                prefixes = {prefixes}
```
benwoo1110: hmm, any reason for using a set here instead of a list or tuple?

dis_snek/client/client.py
AstreaTSS: Oops, yeah, will fix!

dis_snek/client/client.py
AstreaTSS: Not really - just kinda like sets a lot. Though since I'm here committing stuff, I'll just make it a tuple to make it more understandable (the speed difference is too minor to matter).

benwoo1110 (approved):

PR #308, opened 2022-02-21
feat: add `reply` to `MessageContext`

AstreaTSS:

- [ ] Breaking code change

## Description

Going to make this short - to reply to a message when using message commands, you used to have to do `ctx.message.reply`, which was a *bit* annoying when you consider replying to a message command makes a lot of sense and so should be done frequently.
This PR, of course, makes it easier to do so by allowing `ctx.reply`.

## Changes

- Add `reply` to `MessageContext`.
  - This functions the same as `dis_snek.Message.reply`, and thus has the same docstrings and arguments.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/snek/context.py
+++ b/dis_snek/models/snek/context.py
@@ -604,5 +604,15 @@ class MessageContext(Context, SendMixin):
     def content_parameters(self) -> str:
         return self.message.content.removeprefix(f"{self.prefix}{self.invoked_name}").strip()
 
+    async def reply(
+        self,
+        content: Optional[str] = None,
+        embeds: Optional[Union[List[Union["Embed", dict]], Union["Embed", dict]]] = None,
+        embed: Optional[Union["Embed", dict]] = None,
+        **kwargs,
+    ) -> "Message":
+        """Reply to this message, takes all the same attributes as `send`."""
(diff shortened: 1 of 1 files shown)


leestarb (approved):

dis_snek/models/snek/context.py
```diff
@@ -604,5 +604,15 @@ def from_message(cls, client: "Snake", message: "Message"):
     def content_parameters(self) -> str:
         return self.message.content.removeprefix(f"{self.prefix}{self.invoked_name}").strip()
 
+    async def reply(
+        self,
+        content: Optional[str] = None,
+        embeds: Optional[Union[List[Union["Embed", dict]], Union["Embed", dict]]] = None,
+        embed: Optional[Union["Embed", dict]] = None,
```
benwoo1110, line 611: hmm, do you think we should list out all the parameters, to mirror .send()?

AstreaTSS: That's something that's up to another PR, as that would mean `Message.reply` would need to be changed, too. I want to keep this PR rather simple.

It *would* be nice, but it would also be a pain to maintain, so I could argue it either way.

benwoo1110 (approved):

silasary (approved):

PR #316, opened 2022-02-21
 feat: Update debug_exec.py

AlbertUnruh:
- if `exec` returns an iterable containing `Embed`s a `Paginator` will be created from them

- [ ] Breaking code change

## Description

This PR adds the ability to return an iterable with `Embed`s in `exec` and creates a `Paginator` from them which will be send as result.

(An addition to #315)

## Changes

- add check if result is an iterable containing `Embed`s
  - if so a `Paginator` will be created from them and send as result

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/ext/debug_scale/debug_exec.py
+++ b/dis_snek/ext/debug_scale/debug_exec.py
@@ -85,7 +85,7 @@ class DebugExec(Scale):
         if hasattr(result, "__iter__"):
             l_result = list(result)
             if all([isinstance(r, Embed) for r in result]):
-                paginator = Paginator.create_from_embeds(self.bot, *l_results)
+                paginator = Paginator.create_from_embeds(self.bot, *l_result)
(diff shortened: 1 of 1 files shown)


dis_snek/ext/debug_scale/debug_exec.py
```diff
@@ -82,6 +82,13 @@ async def handle_exec_result(self, ctx: MessageContext, result: Any, value: Any)
         if isinstance(result, Paginator):
             return await result.send(ctx)
 
+        if hasattr(result, "__iter__"):
+            l_result = list(result)
+            if all([isinstance(r, Embed) for r in result]):
+                paginator = Paginator.create_from_embeds(self.bot, *l_result)
+                return await paginator.send(ctx)
+            del l_result
```
LordOfPolls: `del l_result` is redundent, garbage collection will handle it

PR #321, opened 2022-02-22
feat: Add `GuildChannel.add_permission` helper method

### LINKED ISSUE #317: [BUG] Permission Overwrites don't work/are poorly documented
**Describe the bug**
As far as I can tell, there is no way to actually use permission overwrites to edit permissions on a channel. Getting existing overwrites returns the overwrite, but there is no way to link the overwrite to a user/role as it currently stands.

**To Reproduce**
```py
# List all overwrites and show the IDs
overwrites = channel._permission_overwrites
overwrites
# Will output false
any(x in guild._role_ids or x in guild._member_ids for x in overwrites.keys())
```

image

**Expected behaviour**
Overwrites should have a clear link to the user/role that they are attached to, as well as a way to actually utilize them in a meaningful way (i.e. `channel.edit_permissions(type=OverwriteTypes.ROLE, target=role, deny=Permissions.READ_MESSAGES)`)

**Additional information**
This isn't so much of a bug as a missing feature, but is also a bug because it is currently impossible to edit permissions afaik.

zevaryx:

- [ ] Breaking code change

## Description

This adds a helper method to `GuildChannel` to add a new permission to a channel (or update an existing permission if it exists). This closes #317, as it makes it much easier to handle permissions in a meaningful way

## Changes

- Add `GuildChannel.add_permission`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/channel.py
+++ b/dis_snek/models/discord/channel.py
@@ -12,6 +12,7 @@ from dis_snek.client.utils.attr_utils import define, field
 from dis_snek.client.utils.converters import optional as optional_c
 from dis_snek.client.utils.converters import timestamp_converter
 from dis_snek.client.utils.input_utils import _bytes_to_base64_data
+from dis_snek.client.utils.misc_utils import get
 from dis_snek.client.utils.serializer import to_dict, to_image_data
 from dis_snek.models.discord.base import DiscordObject
 from dis_snek.models.discord.snowflake import Snowflake_Type, to_snowflake, to_optional_snowflake, SnowflakeObject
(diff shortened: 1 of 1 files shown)


dis_snek/models/discord/channel.py
```diff
@@ -845,6 +846,61 @@ def permissions_for(self, instance: Snowflake_Type) -> Permissions:
...
+
+        permission_overwrites = self.permission_overwrites
+        permission_overwrites.append(overwrite)
+
+        payload = {  # TODO Proper processing
+            "name": self.name,
+            "position": self.position,
+            "permission_overwrites": permission_overwrites,
+        }
+        return await self._edit(payload=payload, reason=reason)
```
benwoo1110: This part needs to change to align with the latest edit method updates.

dis_snek/models/discord/channel.py
```diff
@@ -845,6 +846,61 @@ def permissions_for(self, instance: Snowflake_Type) -> Permissions:
...
+            reason: The reason for this change
+
+        Raises:
+            ValueError: Invalid target for permission
+        """
+        allow = allow or []
+        deny = deny or []
+        if not isinstance(target, PermissionOverwrite):
+            target_type = None
+            if isinstance(target, models.User):
```
benwoo1110:
```suggestion
            if isinstance(target, (models.User, models.Member)):
```

`Member` class should apply also

dis_snek/models/discord/channel.py
```diff
@@ -845,6 +846,61 @@ def permissions_for(self, instance: Snowflake_Type) -> Permissions:
 
             return self.permissions_for(instance)
 
+    async def add_permission(
+        self,
+        target: Union["PermissionOverwrite", "models.Role", "models.User"],
+        type: "OverwriteTypes",
+        allow: Optional[List["Permissions"]] = None,
+        deny: Optional[List["Permissions"]] = None,
```
benwoo1110: I think should have support for a single `int` value passed in.

dis_snek/models/discord/channel.py
```diff
@@ -845,6 +846,61 @@ def permissions_for(self, instance: Snowflake_Type) -> Permissions:
...
+            overwrite = PermissionOverwrite(
+                id=target.id, type=target_type, allow=Permissions.NONE, deny=Permissions.NONE
+            )
+            for perm in allow:
+                overwrite.allow |= perm
+            for perm in deny:
+                overwrite.deny |= perm
+        else:
+            overwrite = target
+        if exists := find(lambda x: x.id == overwrite.id and x.type == overwrite.type, self.permission_overwrites):
```
benwoo1110: hmm do we have a `get_override` method or similar? The lambda looks rather long

zevaryx: I swapped out `find` for `get`, since `get` takes kwargs instead of a lambda

dis_snek/models/discord/channel.py
zevaryx: Added support for ints as well as a list of Permissions

dis_snek/models/discord/channel.py
zevaryx: Updated

dis_snek/models/discord/channel.py
zevaryx: It now calls `self.edit(permission_overwrites=permission_overwrites)`

PR #322, opened 2022-02-22
refactor: Move processing of color from to_dict to a separate method.

benwoo1110:

- [ ] Breaking code change

## Changes

* Add a new `process_color` method.
* Embed color use export converter to `process_color` instead of overriding `to_dict`.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/color.py
+++ b/dis_snek/models/discord/color.py
@@ -1,7 +1,7 @@
 import colorsys
 import re
 from enum import Enum
-from typing import Tuple
+from typing import Tuple, Union
 
 import attr
 
@@ -10,10 +10,12 @@ __all__ = [
     "BrandColors",
     "MaterialColors",
     "FlatUIColors",
+    "process_color",
     "Colour",
     "BrandColours",
     "MaterialColours",
     "FlatUIColours",
+    "process_colour",
 ]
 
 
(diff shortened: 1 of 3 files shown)


dis_snek/models/discord/color.py
```diff
@@ -193,8 +195,22 @@ class FlatUIColors(Color, Enum):
     ASBESTOS = "#7F8C8D"
 
 
+def process_color(color: Union[Color, tuple, list, str, int]) -> int:
```
leestarb, line 198: The dict type is missing

leestarb (approved):

dis_snek/models/discord/color.py
leestarb: The dict type is missing

leestarb (changes requested):

dis_snek/models/discord/color.py
```diff
@@ -193,8 +195,22 @@ class FlatUIColors(Color, Enum):
     ASBESTOS = "#7F8C8D"
 
 
+def process_color(color: Union[Color, dict, tuple, list, str, int]) -> int:
+    if not color:
```
artem30801, line 199: is it for None/Missing? Then it should be reflected in argument type hint

benwoo1110: aaa

PR #335, opened 2022-02-25
docs: A half-decent list of FAQs for porting from dpy

silasary:

- [ ] Breaking code change
- [ ] Tests change

## Description
I put some text on the placeholder page!

## Checklist

- [ ] I've tested my code

### DIFF
--- a/docs/src/Guides/99 Migration From D.py.md
+++ b/docs/src/Guides/99 Migration From D.py.md
@@ -1,4 +1,25 @@
 # Migration From discord.py
 
-!!! danger "Notice"
-    This page is a placeholder, and will be populated at a later date.
+1. Dis-Snek requires python 3.10 (as compared to dpy's 3.5), you may need to upgrade python.
+  - If you see `ERROR: Could not find a version that satisfies the requirement dis_snek (from versions: none)` when trying to `pip install dis-snek`, this is your problem.
+2. Classes/Models
+  - Your client is `dis_snek.Snake`.  (Note that commands are a first-class feature, so this is a replacement to both `discord.Client` and `discord.ext.commands.Bot`)
+  - Cogs are `Scales`.
+  - `Member` is not a subclass of `User`, if you're using `isinstance`, you'll want to check both explicitly.
+3. Scales (Cogs)
+  - These work mostly the same, with a few notable changes:
+  - Your setup function doesn't need to do `bot.add_cog()`.  Simply call `MyCog(bot)`, and it'll automatically register itself.
+  - Scales already define `self.bot`, you don't need to do that in your `__init__` function.
+  - For a full example, see [here](/Guides/20 Scales/)
+4. Event handlers
+  - Register event handlers with `@listen`
+  - Where possible, we use the official names for events, most notably `on_message_create` instead of dpy's `on_message`.
+    - A full list can be found [here](/API Reference/events/discord/).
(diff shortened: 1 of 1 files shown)


docs/src/Guides/99 Migration From D.py.md
```diff
@@ -1,4 +1,25 @@
...
+  - Your client is `dis_snek.Snake`.  (Note that commands are a first-class feature, so this is a replacement to both `discord.Client` and `discord.ext.commands.Bot`)
+  - Cogs are `Scales`.
+  - `Member` is not a subclass of `User`, if you're using `isinstance`, you'll want to check both explicitly.
+3. Scales (Cogs)
+  - These work mostly the same, with a few notable changes:
+  - Your setup function doesn't need to do `bot.add_cog()`.  Simply call `MyCog(bot)`, and it'll automatically register itself.
+  - Scales already define `self.bot`, you don't need to do that in your `__init__` function.
+  - For a full example, see [here](/Guides/20 Scales/)
+4. Event handlers
+  - Register event handlers with `@listen`
```
leestarb: Imo you should rewrite as  `@dis_snek.listen` instead

silasary: I went back and forth on this one. Anyone else have opinions?

AstreaTSS: I think `@dis_snek.listen` is a bit clearer since otherwise it seems rather vague.

docs/src/Guides/99 Migration From D.py.md
```diff
@@ -1,4 +1,25 @@
...
+  - Where possible, we use the official names for events, most notably `on_message_create` instead of dpy's `on_message`.
+    - A full list can be found [here](/API Reference/events/discord/).
+  - Event details are stored on a model, passed as a single parameter. (eg: `on_member_update(before, after)` becomes `on_member_update(event)`, where event has a `.before` and `.after`.
+  - For more details, read [the Events guide](/Guides/10 Events).
+5. Migrating your commands
+  - If you were already using dpy's command extension, migrating to slash commands is fairly simple.  You just need to convert the decorators as per the [Slash Commands guide](/Guides/03 Creating Commands/) (or keep them as message commands with the `message_command` decorator)
```
AstreaTSS: This might create a false expectation that `@message_command` will work somewhat well with the command extension, while in reality it would be a huge rework.

Does `molter` exist? Yeah, but I'm not exactly sure if we're allowed to reference it.
Is this a *very* minor point and might not be worth addressing? Also maybe.

(Also have *no* idea why it selected a huge chunk of it, just focus on the message command part.)

silasary: I'm happy to reference molter, but it seems like a good time to at least move molter into the Snake-Pit org

docs/src/Guides/99 Migration From D.py.md
leestarb: D.py users may think that its `@Snake.listen`

silasary: Yeah, good catch

docs/src/Guides/99 Migration From D.py.md
AstreaTSS: And done!

AstreaTSS (approved):

leestarb (approved):

LordOfPolls (approved):

PR #344, opened 2022-02-28
fix: Adjust typehinting for project packaging

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
#174 was wrongly thought to make sure `.pyi` files were being included when installing `Dis-Snek` from PIP. To an extent, it *did*, as using `pip install .` in the `Dis-Snek` project itself would work but installing it from a wheel or using PyPI would not include these files.

This PR fixes that, making sure those files are included while building the package.

EDIT: as per request, `py.typed` has also been fixed.

## Changes

- Remove `package_data` from `setup.py` and use `include_package_data` instead.
  - Globbing was *not* the way and it seems like `setup.py` ignored it.
- Create `MANIFEST.in` for `include_package_data` and make it include `.pyi` files (and some other basic files).
- EDIT: Move `py.typed` to the correct spot and add it to `MANIFEST.in`.

## Checklist

- [x] I've tested my code

## Additional Note

EDIT: The below note has been adjusted for in this PR now, as per request.

While doing this PR, I found out we *technically* don't have a `py.typed`, which is why it's not included in `MANIFEST.in`. That *probably* needs to be added in a separate PR, though I can add it here if needed.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include README.md
 include LICENSE
 include requirements.txt
+include dis_snek/py.typed
 recursive-include dis_snek *.pyi
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,4 @@
+include README.md
+include LICENSE
+include requirements.txt
+recursive-include dis_snek *.pyi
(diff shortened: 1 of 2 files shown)


silasary:
We do have a py.typed, it's just not in the right place :/
https://github.com/Discord-Snake-Pit/Dis-Snek/blob/master/py.typed

LordOfPolls:
re @silasary and `additional note`

imo that should probs be included in this pr

AstreaTSS:
> re @silasary and `additional note`
> 
> imo that should probs be included in this pr

And included! The PR name (and description) has also been adjusted so that it makes sense.

silasary (approved):

PR #347, opened 2022-03-02
fix: Allow standalone instances to use `guild.fetch_channel`

zevaryx:

- [ ] Breaking code change
- [ ] Tests change

## Description

If a standalone instance is logged in via `await client.login(token)`, running `await guild.fetch_channel(channel_id)` will always return None due to a check to see if `channel_id` is in `guild._channel_ids`. However, this attribute is only populated if you connect to the gateway, and as such, will always fail for standalone instances. 

This PR fixes this by adding an additional check to see if the gateway is started, and running an API check if it is not. This allows standalone instances to properly run `guild.fetch_channel(channel_id)`, instead of having to rely strictly on `client.fetch_channel(channel_id)`

## Changes

- Add extra check to allow standalone instance to use `guild.fetch_channel`

## Checklist

- [x] I've tested my code

## Screenshots
Before:
image

After:
image

### DIFF
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -1230,13 +1230,19 @@ class Guild(BaseGuild):
 
         """
         channel_id = to_snowflake(channel_id)
-        if channel_id in self._channel_ids:
+        if channel_id in self._channel_ids or self._client.start_time is MISSING:
+            # The latter check here is to see if the bot is running with the gateway.
+            # If not, then we need to check the API since only the gateway
+            # populates the channel IDs
+
             # theoretically, this could get any channel the client can see,
             # but to make it less confusing to new programmers,
             # i intentionally check that the guild contains the channel first
             try:
-                return await self._client.fetch_channel(channel_id)
-            except NotFound:
+                channel = await self._client.fetch_channel(channel_id)
+                if channel.guild.id == self.id:
+                    return channel
+            except (NotFound, AttributeError):
                 return None
 
         return None
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -1230,7 +1230,7 @@ class Guild(BaseGuild):
 
         """
         channel_id = to_snowflake(channel_id)
-        if channel_id in self._channel_ids or self._client.start_time is MISSING:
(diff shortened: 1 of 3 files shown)


dis_snek/models/discord/guild.py
```diff
@@ -1230,13 +1230,19 @@ async def fetch_channel(self, channel_id: Snowflake_Type) -> Optional["models.TY
 
         """
         channel_id = to_snowflake(channel_id)
-        if channel_id in self._channel_ids:
+        if channel_id in self._channel_ids or self._client.start_time is MISSING:
```
artem30801: I would rather add a property to client (Snake) like `client.gateway_started` and use this property here in such checks

zevaryx: Resolved in my latest commit

silasary (approved):
It's not elegant, but it works

zevaryx:
I'm 100% open to feedback on how to make this better

dis_snek/models/discord/guild.py
```diff
@@ -1230,13 +1230,19 @@ async def fetch_channel(self, channel_id: Snowflake_Type) -> Optional["models.TY
...
+            # populates the channel IDs
+
             # theoretically, this could get any channel the client can see,
             # but to make it less confusing to new programmers,
             # i intentionally check that the guild contains the channel first
             try:
-                return await self._client.fetch_channel(channel_id)
-            except NotFound:
+                channel = await self._client.fetch_channel(channel_id)
+                if channel.guild.id == self.id:
```
leestarb: `channel._guild_id` (or guild_id not sure)

zevaryx: Changed, forgot about that property

silasary (approved):
I'm happy with this

LordOfPolls:
imo a nicer solution would be to use an asyncio event rather than a boolean. that means you could await the gateway starting up, which opens up far more use cases without losing the functionality you need.

zevaryx:
I changed the `State.gateway_started` property to be an `asyncio.event` object, and `Client.gateway_started` is a boolean that wraps `State.gateway_started.is_set()`

AstreaTSS (approved):

LordOfPolls (approved):

PR #348, opened 2022-03-03
Fix: DebugScale result handling

AlbertUnruh:

- [ ] Breaking code change
- [ ] Tests change

## Description

Fixes the result handling from the exec-cmd.
It only checked for ``bool(result)`` and some models haven't implemented ``__bool__`` which results in ``False`` (besides that the actual result can be a boolean). 

## Changes

- change ``if not result`` to ``if result is None``
- in case of ``isinstance(result, Embed)``: put the result into a list

## Checklist

- [x] I've tested my code


leestarb (approved):
Looks good

silasary (approved):
Awesome

PR #354, opened 2022-03-05
fix: Make channel creation actually add to cache

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
It was found out that new channels made after the bot was started up apparently were not added to the cache. Looking through the code, it turns out that `dis-snek` was actually *removing* the channel from the cache upon a channel's creation by mistake - it was likely caused by copy and pasted code.

Of course, this PR fixes that.

## Changes

- Use `cache.place_channel_data` instead of using custom logic to add the channel to the cache for `_on_raw_channel_create`.
  - *Technically*, this could have been fixed by making `self.cache.channel_cache.pop(channel.id, None)` into `self.cache.channel_cache[channel.id] = channel`, but I decided it was best using methods the cache had for consistency reasons. There isn't a reason why we *shouldn't* use it, after all.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/api/events/processors/channel_events.py
+++ b/dis_snek/api/events/processors/channel_events.py
@@ -19,13 +19,7 @@ log = logging.getLogger(logger_name)
 class ChannelEvents(EventMixinTemplate):
     @Processor.define()
     async def _on_raw_channel_create(self, event: "RawGatewayEvent") -> None:
-
-        channel = BaseChannel.from_dict_factory(event.data, self)
-        if guild := channel.guild:
-            # delete references to this channel in the parent guild
-            guild._channel_ids.add(channel.id)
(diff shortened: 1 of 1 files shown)


benwoo1110 (approved):

benwoo1110:
looks like failing test is due to #355 and unrelated to this pr changes

AstreaTSS:
> looks like failing test is due to #355 and unrelated to this pr changes

Yeah, I actually found that bug *because* of testing this PR, funny enough.

LordOfPolls:
I came expecting an essay, all i got was a short. now im sad

AstreaTSS:
> I came expecting an essay, all i got was a short. now im sad

Oh, you're getting there :)

PR #355, opened 2022-03-06
fix: Fix some bad Type Annotations

silasary:

- [ ] Breaking code change
- [ ] Tests change

## Description
Fixes a a bad annotation in `Channel.add_permission` that was causing the library to not be loaded at all.
Also some corrections to a few decorators.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/channel.py
+++ b/dis_snek/models/discord/channel.py
@@ -894,7 +894,7 @@ class GuildChannel(BaseChannel):
 
     async def add_permission(
         self,
-        target: "PermissionOverwrite" | "models.Role" | "models.User" | "models.Member",
+        target: Union["PermissionOverwrite", "models.Role", "models.User", "models.Member"],
         type: "OverwriteTypes",
         allow: Optional[List["Permissions"] | int] = None,
         deny: Optional[List["Permissions"] | int] = None,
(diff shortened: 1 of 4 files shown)


AstreaTSS (approved):

dis_snek/models/discord/channel.py
```diff
@@ -894,7 +894,7 @@ def permissions_for(self, instance: Snowflake_Type) -> Permissions:
 
     async def add_permission(
         self,
-        target: "PermissionOverwrite" | "models.Role" | "models.User" | "models.Member",
+        target: Union["PermissionOverwrite", "models.Role", "models.User", "models.Member"],
```
AstreaTSS, line 897: Just gonna note that this technically _could_ be written as `target: "PermissionOverwrite | models.Role | models.User | models.Member"` as suggested on the Discord server, but I personally prefer `Union` here anyways.

AstreaTSS (approved):
Uh, ignore GitHub on mobile being a pain to work with, resulting in me reviewing twice.

dis_snek/models/discord/channel.py
silasary: Yeah, I considered it, but IDE integration is generally better with Union if we're using quotes

benwoo1110 (approved):

PR #360, opened 2022-03-07
feat: Add support for Annotated[]

silasary:

- [ ] Breaking code change
- [ ] Tests change

## Description
As recently mentioned in #lib-discussion, there's currently not way to use parameter annotations and also correctly type hint your functions.  Thankfully, [PEP-593](https://www.python.org/dev/peps/pep-0593/) was made for this exact situation, and introduces `typing.Annotated` as a way to include both.

Annotated is a generic annotation, where the first parameter is a type hint, and any subsequent parameters are metadata that can be used by other code.

Here's an example of a message command using `CMD_BODY` with and without type hints:

```python
@message_command()
async def untyped(ctx, body: CMD_BODY):
    ...

@message_command()
async def typed(ctx, body: Annotated[str, CMD_BODY]):
    ...
```

(The same also works with `slash_str_option` and friends, over in slash commands)
```python
@slash_command('sample')
async def my_command_function(ctx: InteractionContext, integer_option: Annotated[int, slash_int_option("Integer Option")]):
    ...
```

## Changes
- Added support for Annotated[] to message_command annotations
- Added support for Annotated[] to slash command options

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/snek/application_commands.py
+++ b/dis_snek/models/snek/application_commands.py
@@ -403,13 +403,12 @@ class SlashCommand(InteractionCommand):
                 for ann in typing.get_args(val.annotation):
                     if isinstance(ann, SlashCommandOption):
                         annotation = ann
-            
+
             if annotation:
                 if not self.options:
                     self.options = []
                 ann.name = name
                 self.options.append(ann)
-
 
         if self.callback is not None:
             if hasattr(self.callback, "options"):
--- a/dis_snek/models/snek/application_commands.py
+++ b/dis_snek/models/snek/application_commands.py
@@ -3,7 +3,7 @@ import inspect
 import logging
 import re
 from enum import IntEnum
-from typing import TYPE_CHECKING, Callable, Coroutine, Dict, List, Union, Optional, Any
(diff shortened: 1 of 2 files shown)


dis_snek/models/snek/application_commands.py
```diff
@@ -4,6 +4,7 @@
 import re
 from enum import IntEnum
 from typing import TYPE_CHECKING, Callable, Coroutine, Dict, List, Union, Optional, Any
+import typing
```
benwoo1110, line 7: Cant we just do `from typing import Annotated ...`

silasary: I can (and probably should) import `Annotated` directly.  But I have a strong dislike of importing functions like `get_origin` and `get_args` by name.

artem30801:
Is there any actual need to have `slash_int_option` etc and not generic `slash_option` in there given that the type can be inferred from typehint anyway?

benwoo1110:
There are individual methods for each type bcu different option type has different parameters, e.g. `slash_int_option` have min_value and max_value while `slash_str_option` won't have such parameters.

benwoo1110 (approved):

PR #365, opened 2022-03-08
feat💥: remove request_fallback args from cache

AstreaTSS:

- [x] Breaking code change
- [ ] Tests change

## Description

`request_fallback` was a method back when the cache had inconsistent/no get and fetch methods - it was an argument used to make `fetch_x` function like `get_x`.

These days, get and fetch methods have been standardized, and there is no need for `request_fallback` to even be a thing - its argument can be accomplished, with certainty, by using `get_x`. However, it was kept to minimize breaking code.

Having legacy code isn't *exactly* the best idea, though. So this PR removes it entirely and adjusts cases where it was used.

## Changes

- `smart_cache.py`:
  - 💥 All `fetch_x` methods have their `request_fallback` argument removed.
  - 💥 The above methods' (typehints *and* actual functions at points) have also been changed so they no longer return `Optional[x]`, instead just returning `x`.
    - This might be controversial for `fetch_role` due to how that works.
  - 💥 Given `is_user_in_guild` and `fetch_user_guild_ids` are asynchronous functions, I decided to make them act like as if the `fallback`s they used to have were `True` and remove those arguments. These aren't methods used anywhere else in the library itself, but if this isn't a desired change, I can always fix it.
- Events and other places that used `request_fallback` have been changed so they use the appropriate get or fetch method instead.

## Checklist

- [x] I've tested my code*
\* Due to how much this PR affects, there's a chance I might have forgotten to test something. I don't think I've missed anything, but feel free to test this out!

### DIFF
--- a/dis_snek/client/smart_cache.py
+++ b/dis_snek/client/smart_cache.py
@@ -84,13 +84,12 @@ class GlobalCache:
 
     # region User cache
 
-    async def fetch_user(self, user_id: "Snowflake_Type", request_fallback: bool = True) -> Optional[User]:
+    async def fetch_user(self, user_id: "Snowflake_Type") -> User:
         """
         Fetch a user by their ID.
 
         Args:
             user_id: The user's ID
-            request_fallback: Should data be requested from Discord if not cached?
 
         Returns:
             User object if found
@@ -99,7 +98,7 @@ class GlobalCache:
         user_id = to_snowflake(user_id)
 
         user = self.user_cache.get(user_id)
-        if request_fallback and user is None:
+        if user is None:
             data = await self._client.http.get_user(user_id)
             user = self.place_user_data(data)
         return user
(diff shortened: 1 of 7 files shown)


dis_snek/client/smart_cache.py
```diff
@@ -350,15 +344,16 @@ def get_user_guild_ids(self, user_id: "Snowflake_Type") -> List["Snowflake_Type"
     # region Message cache
 
     async def fetch_message(
-        self, channel_id: "Snowflake_Type", message_id: "Snowflake_Type", request_fallback: bool = True
+        self,
+        channel_id: "Snowflake_Type",
+        message_id: "Snowflake_Type",
     ) -> Optional[Message]:
```
benwoo1110: Should remove the `Optional` typing here

AstreaTSS: How did I miss that— fixed!

silasary (approved):

benwoo1110 (approved):

PR #370, opened 2022-03-11
Voice Support

LordOfPolls:
Stop asking me when voice; here voice. But it not work good

- [x] Breaking code change
- [ ] Tests change

## Description

Adds voice support to the library, by popular request. 

Example:
```python
@dis_snek.slash_command("play", "play a song!")
@dis_snek.slash_option("song", "The song to play", 3, True)
async def play(self, ctx: dis_snek.InteractionContext, song: str):
    if not ctx.voice_state:
        # connect to the authors voice channel
        await ctx.author.voice.channel.connect()    

    # get the requested song with YTDL
    audio = await YTDLAudio.from_url(song)
    await ctx.send(f"Playing **{audio.entry['title']}**")
    await ctx.voice_state.play(audio)  # play
```

## Changes

- Add required voice components: `voice/`-`audio` `encryption` `opus` `player` `voice_gateway` and `models/snek/active_voice_state`
- Refactor existing gateway to have a parent class with shared methods between `voice_gateway` and `gateway`
- Add helper methods to access a user voice state
- Add helper methods to connect to a vc
- Properly cache guild-creation voice state data
- cache client voice states seperately from the normal voice states
- Refactor build dependecies to go through setup.py
- Add `voice` `all` and `docs` extras 
- Include opus dlls for windows users

## Checklist

- [x] I've tested my code


LordOfPolls:
Rebased PR to incorperate dev changes

LordOfPolls:
Rebased PR to incorperate dev changes

dis_snek/api/gateway/gateway.py
```diff
@@ -407,7 +211,7 @@ async def dispatch_event(self, data, seq, event) -> None:
                 self.sequence = seq
                 self.session_id = data["session_id"]
                 log.info("Connected to gateway!")
-                log.debug(f"Session ID: {self.session_id} Trace: {self._trace}")
+                log.debug(f" Session ID: {self.session_id} Trace: {self._trace}")
```
benwoo1110: _random space?_

dis_snek/models/discord/channel.py
```diff
@@ -1734,6 +1735,16 @@ def voice_members(self) -> List["models.Member"]:
         """Returns a list of members that are currently in the channel. Note: This will not be accurate if the bot was offline while users joined the channel"""
         return [self._client.cache.get_member(self._guild_id, member_id) for member_id in self._voice_member_ids]
 
+    @property
+    def voice_state(self) -> "ActiveVoiceState":
```
benwoo1110:
```suggestion
    def voice_state(self) -> Optional["ActiveVoiceState"]:
```

dis_snek/client/smart_cache.py
```diff
@@ -772,6 +774,41 @@ def delete_voice_state(self, user_id: "Snowflake_Type") -> None:
 
     # endregion Voice cache
 
+    # region Bot Voice cache
+
+    def get_bot_voice_state(self, guild_id: "Snowflake_Type") -> Optional[ActiveVoiceState]:
```
benwoo1110: To standardise with #375, should allow for optional guild_id being passed in

dis_snek/api/voice/player.py
```diff
@@ -0,0 +1,145 @@
...
+from dis_snek.api.voice.opus import Encoder
+
+if TYPE_CHECKING:
+    from dis_snek.models.snek.active_voice_state import ActiveVoiceState
+__all__ = ["Player"]
+
+log = logging.getLogger(logger_name)
+
+
+class Player(threading.Thread):
```
Bluenix2, line 20: Just curious, have you done timing on the voice here? How long does it take to compute (ie. time blocking the event loop)?

LordOfPolls: It wasnt blocking the event loop that was the problem I was solving, it was voice not firing consistently enough when running as an asyncio task. Not to mention, with discord's voice implimentation, you **must** send the payload on time, or you will experience stuttering. As such a threaded solution is prefered over an asyncio task. 

If voice sends too early, you get fast-forward artifacts, if it fires too late, you get stuttering or discord will just outright ignore the packet

zevaryx (approved):
LGTM, have been testing it as changes have been made, and have been following it closely

dis_snek/api/gateway/gateway.py
LordOfPolls: Fixed in 2e7b8741367ffd8559cb5b82e18d853bef53c9f5

dis_snek/client/smart_cache.py
LordOfPolls: Fixed in 6a0c9b4ffe118a4a651a61c4e7fa1e36ddeac9da

deadkex:
The things i understood looked good to me. I also tested this a lot while it got implemented (see "Voice Stuff" thread).

LordOfPolls:
Requesting some additional reviews of this, due to the scope of the PR

LordOfPolls:
Re-synced with dev branch

benwoo1110 (approved):
Tested and fixed all the bugs i found

dis_snek/models/discord/user.py
```diff
@@ -348,6 +349,11 @@ def guild_permissions(self) -> Permissions:
 
         return permissions
 
+    @property
+    def voice(self) -> Optional["VoiceState"]:
```
Kigstn, line 353: This is inconsistent with other usages

Kigstn:
```suggestion
    def voice_state(self) -> Optional["VoiceState"]:
```

dis_snek/client/client.py
```diff
@@ -1831,6 +1832,27 @@ async def fetch_voice_regions(self) -> List["VoiceRegion"]:
...
+            muted: Whether the bot should be muted when connected.
+            deafened: Whether the bot should be deafened when connected.
+
+        Returns:
+            The new active voice state on successfully connection.
+
+        """
+        return await self._connection_state.voice_connect(guild_id, channel_id, muted, deafened)
+
+    def get_bot_voice_state(self, guild_id: "Snowflake_Type") -> ActiveVoiceState:
```
Kigstn: This is missing a docstring.

Kigstn: This should also have a helper method in `Guild`

Kigstn: Wrong return annotation
```suggestion
    def get_bot_voice_state(self, guild_id: "Snowflake_Type") -> Optional[ActiveVoiceState]:
```

docs/src/Guides/22 Voice.md
```diff
@@ -0,0 +1,34 @@
...
+    audio = await YTDLAudio.from_url(song)
+    await ctx.send(f"Now Playing: **{audio.entry['title']}**")
+    # Play the audio
+    await ctx.voice_state.play(audio)
+```
+
+Now just join a voice channel, and type run the "play" slash command with a song of your choice.
+
+Congratulations! You've got a music-bot.
+
```
Kigstn, line 36: This should include a section on how to play your own files, like this f.e.

```suggestion

## Your Own Tunes

If you want to play your own audio, you can do that as well. Just create an `AudioVolume` object and that's it!


```py
audio = await AudioVolume("my_audio_file.mp3")
await ctx.voice_state.play(audio)
```

```

File: dis_snek/api/voice/audio.py [hunk 8]
```diff
@@ -0,0 +1,251 @@
...
+        """
+        Reads frame_size ms of audio from source.
+
+        returns:
+            bytes of audio
+        """
+        ...
+
+
+class Audio(BaseAudio):
```
Kigstn, line 104: IMO this needs to supports for raw bytes to be input

dis_snek/models/discord/guild.py
```diff
@@ -216,6 +216,12 @@ def _process_dict(cls, data: Dict[str, Any], client: "Snake") -> Dict[str, Any]:
 
         if welcome_screen := data.get("welcome_screen"):
             data["welcome_screen"] = GuildWelcome.from_dict(welcome_screen, client)
+
```
Kigstn, line 217: This should be remade into a property that returns dict[user_id, VoiceState]. Currently01, `Guild.voice_states` just returns the raw dict

image

docs/src/Guides/22 Voice.md
```diff
@@ -0,0 +1,34 @@
+# Voice Support
+
+So you want to start playing some 🎵tunes🎶 in voice channels? Well let's get that going for you.
+
+First you're going to want to get the voice dependencies installed:
+```
+pip install dis-snek[voice]
+```
+Now you've got those; let's make a simple play command to get you started.
```
Kigstn, line 12: Inform the users they need to have FFmpeg installed

```suggestion
Also, make sure you have installed [FFmpeg](https://ffmpeg.org/) and made it available in the root directory.

Now you've got those; let's make a simple play command to get you started.
```

Kigstn (changes requested):

dis_snek/api/voice/audio.py
LordOfPolls: Thats a niche use-case, that imo should be handled by the user themselves, not the library. Namely because we wont know anything about the bytestream.

dis_snek/models/discord/guild.py
LordOfPolls: Thats not the fault of this code snippet. Caching a voice state is async, so it made more sense to do it in tasks to prevent bogging down when caching *busy* guilds. 

I agree a property should be made, adding it now

LordOfPolls: Resolved in a nice ick way. 

8baddbf3a39c4bd1a0dbd4da3bd4eae5d79ebfb3

Kigstn (approved):

PR #376, opened 2022-03-14
docs:  Added `typing.BinaryIO` to possible parameters for `guild.create_custom_emoji`

silasary:

- [ ] Breaking code change
- [ ] Tests change

## Description
Adds missing annotation.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/guild.py
+++ b/dis_snek/models/discord/guild.py
@@ -4,7 +4,7 @@ import logging
 import time
 from io import IOBase
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


benwoo1110:
We should add that to all file related type hinting. Maybe a suggestion is to create a variable similar to `Snowflake_Type` so it's less of a pain to update type hinting in the future.

silasary:
Yeah probably.  I'll do a pass of the codebase soon

PR #380, opened 2022-03-15
fix: adjust `define` typehinting

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
As you can clearly see, `define` continues to be a pain in the butt.

Previously, it was thought that all problems related to `define` were solved once and for all, with the function being typehinted as `Optional`, all typos being fixed... but there was just one problem:

Whenever a class that used this decorator and a Visual Studio Code user attempted to use it for their own bot, VSC/pyright would mistaking think the class had 0 expected arguments for its `__init__`. Obviously, this wasn't true, especially with classes like `Embed`.

After an hour of painfully messing around with typehintings, I eventually found out that we were using `__dataclass_transform__` incorrectly, and forgot to typehint the return arguments for `define`, causing issues.
This PR fixes that, of course.
~~oml i hate this decorator~~

## Changes
(All of this happened in the `.pyi` file, *not* the `.py` file.)
- `define` now returns `Callable[[_T], _T]`, solving the argument issue.
- The `f` variable in `define` wasn't needed, so it was removed. The `Optional` import was also removed as `f` was the only thing that used it.
- Move `define`'s definition in the stub *after* `field` to allow it to be used as a field descriptor.
- Adjust the arguments used for `__dataclass_transform__` so that:
  - It reflected `class_defaults`.
  - The `field_descriptors` are now `field` and `attrs.field`. Why was `attrs.define` here?

## Checklist

- [x] I've tested my code


benwoo1110 (approved):

silasary:
Worth noting that this fixes pyright, but not mypy.

AstreaTSS:
> Worth noting that this fixes pyright, but not mypy.

Yeah, `mypy` *does not* like this behavior at all. [This section](https://www.attrs.org/en/stable/extending.html#mypy) of attrs's "Extending" section in their docs details how it can be done, but it is somewhat of a hack.

silasary:
At some point ~~we'll~~ I'll have to write a mypy plugin.  I dread the day I sit down to actually do it.

PR #381, opened 2022-03-15
fix: check for more cases for Role ordering

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
Ordering methods (`>`, `>=`, `<`, `<=`) for `Role` were notoriously buggy, not always correctly determining if a role is higher than another in the role hierarchy.

This turned out to be an issue with how Discord... is, really. There were edge cases we missed.
This PR accounts for that.

## Changes

- Added more cases for `__lt__` in `Role` to check for.
  - It now checks if the role is the `@everyone` role for a guild and if Discord says they have the same position... for some reason. Discord does that at times.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/role.py
+++ b/dis_snek/models/discord/role.py
@@ -58,7 +58,20 @@ class Role(DiscordObject):
         if self._guild_id != other._guild_id:
             raise RuntimeError("Unable to compare Roles from different guilds.")
 
-        return self.position < other.position
+        if self.id == self._guild_id:  # everyone role
+            # everyone role is on the bottom, so check if the other role is, well, not it
+            # because then it must be higher than it
+            return other.id != self.id
+
+        if self.position < other.position:
+            return True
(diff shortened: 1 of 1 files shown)


leestarb:
>  seems like "the class should supply an __eq__() method." Not having this method might have screwed with it, so I added it in to make total_ordering happy.

Discord object already has it

AstreaTSS:
> > seems like "the class should supply an **eq**() method." Not having this method might have screwed with it, so I added it in to make total_ordering happy.
> 
> Discord object already has it

...right, it does.

...thankfully I actually bothered to make two reverts, haha. Will revert in like a couple of minutes.

AstreaTSS:
And done. PR name and description adjusted too.

leestarb (approved):

LordOfPolls (approved):

PR #384, opened 2022-03-15
refactor💥: Simplify thread create into one method

benwoo1110:

- [x] Breaking code change
- [ ] Tests change

## Description

Mainly made this change bcu `create_thread_without_message` doesn't work for the news channel. So I think its better to condense to one method and add checks/docstring to inform users on allowed/disallowed behaviours.

## Changes

* Simplify thread create into one method

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/channel.py
+++ b/dis_snek/models/discord/channel.py
@@ -428,61 +428,45 @@ class InvitableMixin:
 
 @define(slots=False)
 class ThreadableMixin:
-    async def create_thread_with_message(
+    async def create_thread(
         self,
         name: str,
-        message: Union[Snowflake_Type, "models.Message"],
-        auto_archive_duration: Union[AutoArchiveDuration, int] = AutoArchiveDuration.ONE_DAY,
-        reason: Optional[str] = None,
-    ) -> Union["GuildNewsThread", "GuildPublicThread"]:
+        message: Absent[Union[Snowflake_Type]] = MISSING,
+        thread_type: Absent[Union[ChannelTypes]] = MISSING,
+        invitable: Absent[bool] = MISSING,
+        auto_archive_duration: Union[AutoArchiveDuration] = AutoArchiveDuration.ONE_DAY,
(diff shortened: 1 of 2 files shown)


dis_snek/models/discord/channel.py
```diff
@@ -428,61 +428,45 @@ async def fetch_invites(self) -> List["models.Invite"]:
...
-    ) -> Union["GuildNewsThread", "GuildPublicThread"]:
+        message: Absent[Union[Snowflake_Type]] = MISSING,
+        thread_type: Absent[Union[ChannelTypes]] = MISSING,
+        invitable: Absent[bool] = MISSING,
+        auto_archive_duration: Union[AutoArchiveDuration] = AutoArchiveDuration.ONE_DAY,
+        reason: Absent[str] = None,
+    ) -> Union["GuildNewsThread", "GuildPublicThread", "GuildPrivateThread"]:
         """
-        Create a thread connected to a message.
+        Creates a nee thread in this channel. If a message is provided, it will be used as the initial message.
```
LordOfPolls, line 441:
```suggestion
        Creates a new thread in this channel. If a message is provided, it will be used as the initial message.
```

LordOfPolls (commented):
You could have made this non-breaking by leaving the exisitng methods 

suggestion below unrelated

benwoo1110:
> You could have made this non-breaking by leaving the exisitng methods

hm yea makes sense, but `create_thread_without_message` will basically always fail for a guild news channel hmm

LordOfPolls:
```python
if isinstance(channel, GuildNews):
    raise SomeException("Guild News Channels do not support this method")
```

artem30801:
I like how new API looks more than how old API looks. I'm fine with it being breaking as it makes code cleaner too. Better now than later.

artem30801 (approved):

silasary:
> Better now than later.

My personal preference is always "Later, but with Depreciation Warnings."

PR #385, opened 2022-03-15
feat: use overloads to better type get_context

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
While `Snake.get_context`'s typehinting worked, it was quite imprecise, to put it in words. For example, the `get_context` in both `_dispatch_msg_commamnd` and `_dispatch_interaction` would cause issues with typehinting later down in those two functions, with them respectively not expecting fields/types from `InteractionContext` or `MessageContext`.

This PR *somewhat* fixes that. Due to the fact that the interaction data itself isn't really typehinted by type, it can only do so much, but it sure does produce a lot less red.

## Changes

- Use `typing.overload` and interaction data typings from `discord_typings` to typehint `get_context` depending on what data was receieved.
  - A "catch-all" `dict` overload was also added for the sake of `dis-snek`'s internal functions.
  - This involved importing `overload` and `Literal` from `typing`, making `black` trigger and re-format the imports.
  - This does not support modal response interaction typehinting as `discord_typings` does not have that yet. A comment detailing how the overload could be added in the future was left for any future maintainer who needs to add it in.
- Adjust typehintings of the main `get_context` to reflect the overloads.
  - `ModalContext` is now included in the union of types returned by it since, well, why wasn't it included before?

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -9,7 +9,26 @@ import time
 import traceback
 from collections.abc import Iterable
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Callable, Coroutine, Dict, List, NoReturn, Optional, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Coroutine,
+    Dict,
+    List,
+    NoReturn,
+    Optional,
+    Type,
+    Union,
+    overload,
+    Literal,
+)
+
+from discord_typings.interactions.receiving import (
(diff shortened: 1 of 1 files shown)


benwoo1110 (approved):
Thinking of it, there are probably other places that we need overload as well

AstreaTSS:
> Thinking of it, there are probably other places that we need overload as well

That's what I'm thinking - `get_context` was simply the first that came to mind.

silasary (approved):

PR #393, opened 2022-03-16
ci: Make GHA run pre-commit, so we don't have to worry about drift

silasary:
## What type of pull request is this?

- [ ] Non-breaking code change
- [ ] Breaking code change
- [ ] Documentation change/addition
- [x] Tests change

## Description
Modifies Github Actions to run `pre-commit run --all-files`, which should ensure we don't have drift when we inevitably add new flake8 linters (this is foreshadowing)

## Checklist

- [ ] I've formatted my code with [Black](https://black.readthedocs.io/en/stable/)
- [ ] I've ensured my code works on `Python 3.10.x`
- [x] I've tested my code

PR #396, opened 2022-03-16
Revert "ci: Make GHA run pre-commit, so we don't have to worry about drift"

benwoo1110:
Reverts Discord-Snake-Pit/Dis-Snek#393


silasary:
PR does not need to be reverted (as per conversation in #contributors)

PR #407, opened 2022-03-21
feat: add `Paginator.reply`

AlbertUnruh:

- [ ] Breaking code change
- [ ] Tests change

## Description

Adds the ability to reply to a (Message)Context with a Paginator.

## Changes

- add `Paginator.reply`

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/ext/paginators.py
+++ b/dis_snek/ext/paginators.py
@@ -15,6 +15,7 @@ from dis_snek import (
     ComponentCommand,
     get_components_ids,
     Context,
+    MessageContext,
     Message,
     MISSING,
     Snowflake_Type,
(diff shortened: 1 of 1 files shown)


dis_snek/ext/paginators.py
```diff
@@ -286,6 +287,26 @@ async def send(self, ctx: Context) -> Message:
...
+        Returns:
+            The resulting message
+
+        """
+        self._message = await ctx.reply(**self.to_dict())
+        self._author_id = ctx.author.id
+
+        if self.timeout_interval > 1:
+            self._timeout_task = Timeout(self)
+            self.client.loop.create_task(self._timeout_task())
```
LordOfPolls, line 306: It looks like you created this using an older version of dis_snek. `client.loop` no longer exits. I suggest updating your branch, and using `asyncio.create_task` instead.

AlbertUnruh: > It looks like you created this using an older version of dis_snek. `client.loop` no longer exits. I suggest updating your branch, and using `asyncio.create_task` instead.

does it work if I just edit the line?

I accidentally based this PR on `main` and not on `dev`.
Otherwise I would reopen this PR but with the code based on `dev`.

LordOfPolls: The latter would be best, tbh

AlbertUnruh: Ok, I'll close the PR and reopen it in the evening when I'm home (UTC+1 for context)

PR #423, opened 2022-04-04
feat: Allow passing keyword arguments into extension methods

leestarb:
## What type of pull request is this?

- [x] Non-breaking code change
- [ ] Breaking code change
- [ ] Documentation change/addition
- [ ] Tests change

## Description

It's an useful ability!

## Changes

- Add **kwargs to extension and scale methods
- Add docstrings

## Checklist

- [x] I've formatted my code with [Black](https://black.readthedocs.io/en/stable/)
- [x] I've ensured my code works on `Python 3.10.x`
- [x] I've tested my code

PR #424, opened 2022-04-05
docs: Add guides example about #423

leestarb:

- [ ] Breaking code change
- [ ] Tests change

## Description

!

## Changes

- Add "Scale switching" guides example

## Checklist

- [ ] I've tested my code

### DIFF
--- a/docs/src/Guides/20 Scales.md
+++ b/docs/src/Guides/20 Scales.md
@@ -208,7 +208,7 @@ To do this, simply run `bot.regrow_scale("Filename_here")` and your new code wil
 is doing will be abruptly stopped.
 
 
-Finally, you can also pass any **keyword** arguments into module's `setup` and `teardown` functions, it will also send them into your `Scale.shed` method.
+Any keyword-arguments you pass to the `setup` or `teardown` methods, will also be passed to the `Scale.shed` method. 
 
 Here is a basic "Scale switching" example:
 
--- a/docs/src/Guides/20 Scales.md
+++ b/docs/src/Guides/20 Scales.md
@@ -208,7 +208,7 @@ To do this, simply run `bot.regrow_scale("Filename_here")` and your new code wil
 is doing will be abruptly stopped.
 
 
-Any keyword-arguments you pass to the `setup` or `teardown` methods, will also be passed to the `Scale.shed` method. 
+Any keyword-arguments you pass to the `setup` or `teardown` methods, will also be passed to the `Scale.shed` method.
 
 Here is a basic "Scale switching" example:
 
--- a/docs/src/Guides/20 Scales.md
+++ b/docs/src/Guides/20 Scales.md
@@ -203,7 +203,14 @@ bot.grow_scale("Filename_here")
 bot.start("token")
 ```
 
-You can also pass any **keyword** arguments into the `setup` function. Here is a basic "Scale switching" example:
+Additionally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
+To do this, simply run `bot.regrow_scale("Filename_here")` and your new code will be used. Bare in mind any tasks your scale
+is doing will be abruptly stopped.
+
+
+Finally, you can also pass any **keyword** arguments into module's `setup` and `teardown` functions, it will also send them into your `Scale.shed` method.
+
+Here is a basic "Scale switching" example:
 
 ```python
 from dis_snek import Scale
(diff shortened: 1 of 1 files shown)


docs/src/Guides/20 Scales.md
```diff
@@ -203,6 +203,33 @@ bot.grow_scale("Filename_here")
 bot.start("token")
 ```
 
+Use also can pass any **keyword** arguments into the `setup` function. Here is a basic "Scale switching" example:
```
LordOfPolls: Grammatical Error. Do you mean `you can also pass` or `users can also pass`?

LordOfPolls:
Bit nit-picky, but i think these docs should go *after* the docs for reloading, as this is inhernatly more advanced than basic load // unload

leestarb:
> Bit nit-picky, but i think these docs should go *after* the docs for reloading, as this is inhernatly more advanced than basic load // unload 

Maybe will do it someday

But ok, now

LordOfPolls (approved):

zevaryx (approved):
LGtm

docs/src/Guides/20 Scales.md
```diff
@@ -203,6 +203,36 @@ bot.grow_scale("Filename_here")
 bot.start("token")
 ```
 
-Finally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
+Additionally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
```
AstreaTSS: "Additionally" here is... confusing. I think "Now" or something similar would be better.

Nitpick though.

docs/src/Guides/20 Scales.md
```diff
@@ -203,6 +203,36 @@ bot.grow_scale("Filename_here")
...
+
+```python
+from dis_snek import Scale
+
+class SomeScale(Scale):
+    def __init__(self, bot, some_arg: int = 0):
+        ...
+
+class AnotherScale(Scale):
+    def __init__(self, _, another_arg: float = 0.0):
```
silasary: Why are you using `_` instead of bot here?

LordOfPolls: according to pep this would actually be correct, seeing as you dont use the argument, its purely there because `__new__` requires it to be passed. Though it could be argued as the below code is literally an ellipsis placeholder the true argument name should be used

silasary: For the sake of a readable example, we should definitely use the name.

docs/src/Guides/20 Scales.md
```diff
@@ -203,6 +203,36 @@ bot.grow_scale("Filename_here")
 bot.start("token")
 ```
 
-Finally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
+Additionally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
 To do this, simply run `bot.regrow_scale("Filename_here")` and your new code will be used. Bare in mind any tasks your scale
 is doing will be abruptly stopped.
+
+
+Finally, you can also pass any **keyword** arguments into module's `setup` and `teardown` functions, it will also send them into your `Scale.shed` method.
```
AstreaTSS: Am I the only one who got a bit confused by this? I'm *genuinely* asking by the way, it took me a bit to understand but I think that's just me.

LordOfPolls: It's grammatically fragmented. It could be rewritten with clearer wording

silasary: I understand it, but it's definitely not clean

leestarb: How to make it clearer?

LordOfPolls:
```suggestion
Any keyword-arguments you pass to the `setup` or `teardown` methods, will also be passed to the `Scale.shed` method. 
```

LordOfPolls: My suggestion: 
https://github.com/Discord-Snake-Pit/Dis-Snek/pull/424#discussion_r845777544

docs/src/Guides/20 Scales.md
```diff
@@ -203,12 +203,13 @@ bot.grow_scale("Filename_here")
 bot.start("token")
 ```
 
-Additionally, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
+Nowadays, for the cool bit of Scales, reloading. Scales allow you to edit your code, and reload it, without restarting the bot.
```
LordOfPolls: what?

LordOfPolls: I dont think this means what you think it means

LordOfPolls (approved):

PR #427, opened 2022-04-06
feat: Send error to channel with default error handlers

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

The default error handler now sends the error to the channel the command was ran in

## Changes

- Added send call to error handler
- Organised imports and converted relative imports 

## Checklist

- [ ] I've tested my code

### DIFF
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -488,9 +488,32 @@ class Snake(
         """
         await self.on_error(f"cmd /`{ctx.invoked_name}`", error, *args, **kwargs)
         try:
-            out = "".join(traceback.format_exception(error))
-            out = out.replace(self.http.token, "[REDACTED TOKEN]")
-            if self.send_command_tracebacks:
+            if isinstance(error, errors.CommandOnCooldown):
+                await ctx.send(
+                    embeds=Embed(
+                        description=f"This command is on cooldown!\n"
+                        f"Please try again in {int(error.cooldown.get_cooldown_time())} seconds",
+                        color=BrandColors.FUCHSIA,
+                    )
+                )
+            elif isinstance(error, errors.MaxConcurrencyReached):
+                await ctx.send(
+                    embeds=Embed(
+                        description="This command has reached its maximum concurrent usage!\n"
+                        "Please try again shortly.",
+                        color=BrandColors.FUCHSIA,
+                    )
+                )
+            elif isinstance(error, errors.CommandCheckFailure):
+                await ctx.send(
+                    embeds=Embed(
+                        description="You do not have permission to run this command!",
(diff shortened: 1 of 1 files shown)


LordOfPolls:
image

AstreaTSS:
Couldn't this have potentially dangerous information, like bot tokens or API keys? I think it's a bit unwise to show off the full traceback.

LordOfPolls:
Let them leak dammit \s 

Jokes aside, the trace back will leak the token/api key *if* the user was daft enough to hardcode their token/keys in their src files. *Not* showing tracebacks limits the amount of usable information that can be gleamed from the error message. Bot developers could implement their own error messages that allow for looking up errors, but the base lib cant have that functionality. 

I do honestly think that vulnerability is something the bot developer should mitigate, not us. I'd also be inclined to point out other libraries send tracebacks by default too

LordOfPolls:
As the bare-minimum ill filter the bot token. I will make it clear to those reading this 

**DO NOT RELY ON THIS TO KEEP YOUR TOKENS SAFE**
Do the bare minimum security on your projects and put the token in a text file or something

leestarb:
I suggest moving this to the debug scale
+
I dont think this is really useful

dis_snek/client/client.py
```diff
@@ -478,7 +481,17 @@ async def on_command_error(self, ctx: Context, error: Exception, *args, **kwargs
         Override this to change error handling behavior
 
         """
-        return await self.on_error(f"cmd /`{ctx.invoked_name}`", error, *args, **kwargs)
+        await self.on_error(f"cmd /`{ctx.invoked_name}`", error, *args, **kwargs)
+        try:
+            out = "".join(traceback.format_exception(error))
```
leestarb: This needs the [:1990] slice ig

leestarb (changes requested):

LordOfPolls:
> I suggest moving this to the debug scale + I dont think this is really useful

I disagree on both counts

dis_snek/client/client.py
LordOfPolls: Agreed, adding now

LordOfPolls:
image

LordOfPolls:
Requesting reviews

leestarb:
I suggest moving this logic to the utils package

LordOfPolls:
Moving which logic? The logic for the 3 different embeds? 

Is that really utility worthy? It's 3 if statements and some embeds. Surely if youre implimenting your own error handler you'd want to create your own embeds, which would defeat the point of making the logic a util

silasary (approved):

PR #434, opened 2022-04-08
feat: Make logger a variable users can change for errors

Kigstn:

- [ ] Breaking code change
- [ ] Tests change

## Description

It has been annoying me for a while that the standard dis-snek error logging is really good but inflexible. 

Personally, I want the option to have different loggers log different logs which was not really possible before since they all used the internal logger.
This PR does not change this, but it allows users to overwrite the logger and pass their own custom logger.

Example of new functionality:

```py
class MySnake(Snake):
    async def on_command_error(self, ctx: Context, error: Exception, logger: Absent[logging.Logger] = MISSING, *args, **kwargs) -> None:
        logger = logging.getLogger("command_error_logger")
        await super().on_command_error(ctx, error, logger, *args, **kwargs)

    async def on_component_error(self, ctx: ComponentContext, error: Exception, logger: Absent[logging.Logger] = MISSING, *args, **kwargs) -> None:
        logger = logging.getLogger("component_error_logger")
        await super().on_component_error(ctx, error, logger, *args, **kwargs)
```

This way, users can use their own loggers while utilising dis-snek really good error parser / formatter.

## Changes

- make logger a variable users can change for errors

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -435,15 +435,19 @@ class Snake(
         return asyncio.create_task(wrapped, name=f"snake:: {event.resolved_name}")
 
     @staticmethod
-    def default_error_handler(source: str, error: BaseException) -> None:
+    def default_error_handler(source: str, error: BaseException, logger: Absent[logging.Logger] = MISSING) -> None:
         """
         The default error logging behaviour.
 
         Args:
             source: The source of this error
             error: The exception itself
+            logger: The logger to use. Default: Internal Logger
 
         """
+        if logger is MISSING:
+            logger = log
+
         out = traceback.format_exception(error)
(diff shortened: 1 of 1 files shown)


LordOfPolls:
Why would you do this instead of just overriding the apropriate methods?

Kigstn:
> Why would you do this instead of just overriding the apropriate methods?

Simply to avoid having to re-implement the behaviour from `default_error_handler()`. I agree it is a minor thing tho

benwoo1110:
I feel overriding the methods is a better option

LordOfPolls:
Re the poll in the discord server, this is being rejected

PR #441, opened 2022-04-14
feat: support multi-scale unloading

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

cc: https://canary.discord.com/channels/870046872864165888/964080046153281586/964085134934552627
Allows for multiple scales to be defined in one file and successfully unload them. IE

```python
from dis_snek import Scale, message_command


class ScaleOne(Scale):
    @message_command()
    async def scale_one_test(self, ctx):
        await ctx.send(ctx.invoked_name)


class ScaleTwo(Scale):
    @message_command()
    async def scale_two_test(self, ctx):
        await ctx.send(ctx.invoked_name)


def setup(bot):
    ScaleOne(bot)
    ScaleTwo(bot)
```
Previously, reloading the above extension would result in a duplicate command exception, this fixes that. 

## Changes

- Added `Snake.get_scales` which returns a list of all matching scales
- `Snake.unload_extension` now handles the above, and sheds each scale

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client/client.py
+++ b/dis_snek/client/client.py
@@ -1504,22 +1504,24 @@ class Snake(
     async def _disconnect(self) -> None:
         self._ready.clear()
 
-    def get_scale(self, name) -> Optional[Scale]:
+    def get_scale(self, name) -> list[Scale]:
         """
-        Get a scale.
+        Get all scales with a name or extension name.
 
         Args:
             name: The name of the scale, or the name of it's extension
 
         Returns:
-            Scale or None if no scale is found
+            List of Scales
         """
+        out = []
         if name not in self.scales.keys():
             for scale in self.scales.values():
                 if scale.extension_name == name:
-                    return scale
+                    out.append(scale)
+            return out
 
-        return self.scales.get(name, None)
+        return [self.scales.get(name, None)]
 
     def grow_scale(self, file_name: str, package: str = None, **load_kwargs) -> None:
         """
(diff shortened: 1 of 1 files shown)


LordOfPolls:
i'd be remissed if i didnt point out that if your scale usage is too abstract for the builtin unloader, use the `teardown` function to unload the scales

dis_snek/client/client.py
```diff
@@ -1504,22 +1504,24 @@ async def _dispatch_msg_commands(self, event: MessageCreate) -> None:
     async def _disconnect(self) -> None:
         self._ready.clear()
 
-    def get_scale(self, name) -> Optional[Scale]:
+    def get_scale(self, name) -> list[Scale]:
```
AstreaTSS: Shouldn't this be renamed `get_scales` then?

LordOfPolls: i almost did, but i was conflicted about the technically unnecessary rename

AstreaTSS: I mean, this *is* a breaking change already....`get_scale` also doesn't make sense when, well, there can be multiple scales.

LordOfPolls: reasonable, resolved

silasary (commented):
I'd prefer if get_scale remained, even if it's more limited in scope compared to get_scales.

When users are calling it (outside of the reloading use-case), they generally pass the scale name and expect an optional scale as a return value. 
I personally don't want to have to unpack a list when doing inter-scale references

LordOfPolls:
@silasary restored in ae566fae0b609e176617eb901cd51e30e2ebd122

silasary (approved):
Awesome. This is no longer breaking btw

LordOfPolls:
Yeah i realised, just hadnt edited the pr yet

PR #445, opened 2022-04-16
feat💥: Localisation support

### LINKED ISSUE #425: [FEAT] Application Command Localization
**Is your feature request related to a problem? Please describe.**
Discord has started to let us test application command localization (slash and context menus)

**Describe the solution you'd like**
We need an intuitive and efficient way for a developer to provide localization for their application commands.

**Additional information**
For anyone looking to attempt this, here is the build override (just paste in discord and click):
https://discord.com/__development/link?s=GBxWH9lkrFwWI3HEiYg%2FshrD8kLcqS2NHnJn0rgceNU%3D.eyJ0YXJnZXRCdWlsZE92ZXJyaWRlIjp7ImRpc2NvcmRfd2ViIjp7InR5cGUiOiJicmFuY2giLCJpZCI6ImZlYXR1cmUvd2ViLXNsYXNoLWNvbW1hbmQtbG9jYWxpemF0aW9uIn19LCJyZWxlYXNlQ2hhbm5lbCI6bnVsbCwidmFsaWRGb3JVc2VySWRzIjpbXSwiYWxsb3dMb2dnZWRPdXQiOmZhbHNlLCJleHBpcmVzQXQiOiJXZWQsIDA2IEFwciAyMDIyIDE4OjI2OjUyIEdNVCJ9

And here are the docs: https://discord.com/developers/docs/interactions/application-commands#localization

LordOfPolls:

- [x] Breaking code change
- [ ] Tests change

## Description

Breaking in the event someone was performing operations on the name or description fields of app_cmds and their affiliated objects, or using `ctx.invoked_name` in a situation where localisation is not desirable.

Adds support for application command localisation. This is done by converting `name` and `description` fields into a subclass of a new object called `LocalisedField`.
These subclassed objects include all the validation that used to be done in a per-object basis. The syntax looks like this:

```python
@slash_command(name=LocalisedName(english_us="test_cmd", french="la_test_cmd", german="der_test_cmd"))
async def localised_test(ctx: InteractionContext):
    await ctx.send(f"{ctx.invoked_name} was invoked under {ctx.locale}")
````

Syntax for people who don't want to localise is unaffected. The below snippet will work as normal, however localised fields will be converted transparently from string to `LocalisedField`
```python
@slash_command(name="test_command")
async def normal(ctx: InteractionContext):
    await ctx.send(f"{ctx.invoked_name} was invoked under {ctx.locale}")
```

This PR also makes `ctx.invoked_name` localised: meaning it now returns the localised name of the application command. To handle use cases where the non-localised name was needed, `ctx.invoke_target` was added to handle that. This could have been done without being breaking, however the naming would not have made sense. Invoked name sounds like the command the user invoked, which was the original purpose.

## Changes

- Add `LocalisedField` in `models/snek/localisation.py`
- Add `LocalisedName` and `LocalisedDesc` in `models/snek/application_commands.py`
- Change all existing name/desc fields to use `LocalisedFields`
- Update syncing to support localisation

## Checklist

- [x] I've tested my code

### DIFF
--- a/docs/src/Guides/23 Localisation.md
+++ b/docs/src/Guides/23 Localisation.md
@@ -35,6 +35,8 @@ async def hello_cmd(ctx: dis_snek.InteractionContext):
     await ctx.send(f"{ctx.invoked_name} {ctx.author.display_name}")
 ```
 Simply by changing `"hello"` to `ctx.invoked_name` the command will always use whatever the user typed to greet them.
+If you want to know what locale the user is in, simply use `ctx.locale`.
+
 
 This will work for any object with a `name` or `description` field. Simply use `LocalisedDesc` instead for descriptions.
 For example, you can localise options, choices, and subcommands.
--- a/docs/src/Guides/23 Localisation.md
+++ b/docs/src/Guides/23 Localisation.md
@@ -1,0 +1,40 @@
+# Localising
+
+So your bot has grown, and now you need to ~~localize~~ localise your bot. Well thank god we support localisation then, huh?
+
+To clarify; localisation is a feature of application commands that discord offers,
+this means the same command will have different names and descriptions depending on the user's locale settings.
+
+# How its made:
+
+Let's take this nice and simple `hello` command
+```python
+import dis_snek
+
+@dis_snek.slash_command(name="hello")
+async def hello_cmd(ctx: dis_snek.InteractionContext):
+    await ctx.send(f"Hello {ctx.author.display_name}")
+```
+This command was immensely popular, and now we have some 🇫🇷 French users. Wouldn't it be nice if we could speak their language.
+```python
+import dis_snek
+from dis_snek import LocalisedName
+
+@dis_snek.slash_command(name=LocalisedName(english_us="hello", french="salut"))
+async def hello_cmd(ctx: dis_snek.InteractionContext):
+    await ctx.send(f"Hello {ctx.author.display_name}")
+```
+All we need to do is set the field to a `Localised` object, and dis-snek and discord wil handle the rest for you.
+For extra flavour lets make this command more dynamic.
+```python
+import dis_snek
+from dis_snek import LocalisedName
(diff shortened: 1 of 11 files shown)


Kigstn (changes requested):
Looks good, however, I think a small section in the guides should be added on how to use this

zevaryx (approved):
Already partially reviewed this in the Discord, but looks good to me. My only suggestion would be maybe an alias of `dis_snek.models.snek.localization.LocalizedField` before people start saying that it's spelled incorrectly, but no where near a priority, or even remotely blocking

LordOfPolls:
Both of the above comments should not be resolved - please re-review

docs/src/Guides/23 Localisation.md
```diff
@@ -0,0 +1,40 @@
...
+For extra flavour lets make this command more dynamic.
+```python
+import dis_snek
+from dis_snek import LocalisedName
+
+@dis_snek.slash_command(name=LocalisedName(english_us="hello", french="salut"))
+async def hello_cmd(ctx: dis_snek.InteractionContext):
+    await ctx.send(f"{ctx.invoked_name} {ctx.author.display_name}")
+```
+Simply by changing `"hello"` to `ctx.invoked_name` the command will always use whatever the user typed to greet them.
```
silasary, line 37: Is there any way to get the locale from the context, rather than having to cross reference command names?

LordOfPolls: `ctx.locale`

silasary: That should definitely be in the guide

LordOfPolls: Added in 20ffff9eeb699969bc04c2a749546a1138f05e29

silasary (approved):

Kigstn (approved):

PR #449, opened 2022-04-18
Fix: `VoiceChannel.voice_members` was not updating correctly

Kigstn:

- [ ] Breaking code change
- [ ] Tests change

## Description

`VoiceChannel.voice_members` was not updating correctly when people left a voice channel. The user was not correctly removed from their old voice channel. This was due to a copied list that needed to be deepcopied, as far as my tests could tell.

Test Code:

```py
@listen()
async def _event(event: VoiceStateUpdate):
    before = event.before.channel if event.before else None
    before_members = before.voice_members if before else None
    before_members_cache = event.bot.cache.get_channel(before.id).voice_members if before else None

    after = event.after.channel
    after_members = after.voice_members if after else None
    after_members_cache = event.bot.cache.get_channel(after.id).voice_members if after else None

    assert after_members == after_members_cache
    if before:
        assert before_members != before_members_cache
```

## Changes

- deecopy the `_voice_member_ids` instead of copying them
- remove some unnecessary code in place_voice_state_data - this was necessary before the cache rewrite polls did a while ago

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/client/smart_cache.py
+++ b/dis_snek/client/smart_cache.py
@@ -740,12 +740,9 @@ class GlobalCache:
         old_state = self.get_voice_state(user_id)
         if old_state:
             # noinspection PyProtectedMember
-            # for some unholy reason, self.get_channel() seems to return a copy(), and does not update the cache after changing the obj
-            old_channel = self.channel_cache[old_state._channel_id]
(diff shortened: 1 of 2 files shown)


deadkex:
Tested and works

PR #453, opened 2022-04-20
fix: voice members not updating

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

While testing voice states I noticed some odd behaviour; users were not being removed from the voice channel's member list. At first i couldnt work out why, until i found this codeblock in voicestate. I dont understand why its necessary, nor why it was added in the first place; however removing it fixes all the issues ive found with `GuildVoice.voice_members`

## Changes

- Remove weird code 

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/discord/voice_state.py
+++ b/dis_snek/models/discord/voice_state.py
@@ -1,4 +1,3 @@
-import copy
 from typing import TYPE_CHECKING, Optional, Dict, Any
 
 from dis_snek.client.const import MISSING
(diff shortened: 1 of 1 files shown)


zevaryx (approved):
Helped test this fix, can confirm that it works

LordOfPolls:
req review from @Kigstn as they added the code. and @zevaryx as i was talking to her when i found the bug

Kigstn:
Well the problem I had, which led to the janky solution I made, was that when the voice state changes, the user gets removed from the voice channel members list in the cache. 

This results in the problem, that the `event.before` arg is wrong. It does not show the member in the `voice_members` list, since `before.channel` got the channel obj with the updated member list from the cache.
Removing the code block will lead to `event.before.channel.voice_members` to show the `voice_members` for the channel after the event, so without the user who switched channel / disconnected and thus be incorrect.

I dislike the current solution as well, but can't think of a better way to do that while keeping the cache accurate

Kigstn (changes requested):
Well the problem I had, which led to the janky solution I made, was that when the voice state changes, the user gets removed from the voice channel members list in the cache. 

This results in the problem, that the `event.before` arg is wrong. It does not show the member in the `voice_members` list, since `before.channel` got the channel obj with the updated member list from the cache.
Removing the code block will lead to `event.before.channel.voice_members` to show the `voice_members` for the channel after the event, so without the user who switched channel / disconnected and thus be incorrect.

I dislike the current solution as well, but can't think of a better way to do that while keeping the cache accurate

LordOfPolls:
The issue is, this solution renders the `after` payload and the `channel` object itself inaccurate. So by preserving accuracy of the `before` object, it destroys the usefulness of 2 others. Im working on an alternative now - as per our discussion last night

LordOfPolls:
Im trying to replicate with the same code this morning, and i can no longer replicate. I am very confused. The only variable thats changed is discord. I'm going to close this PR until such a time i manage to replicate, or more information is found

PR #458, opened 2022-04-28
feat💥: Add support for perms2.0 and retire support for perms1.0

Kigstn:

- [x] Breaking code change
- [ ] Tests change

## Description

This PR adds support for the permission 2.0 system and retires support for the old system. Testing would be greatly appreciated.

Note for testing: Since discord does not display the default permissions data anywhere I could find (_sigh_), you basically need to inspect the data they send in the synching logic.

## Changes

- BREAKING: remove `default_permission` attribute
- BREAKING: remove `permissions` attribute
- BREAKING: remove `Permission` class
- BREAKING: remove `PermissionTypes` class
- BREAKING: remove `slash_permission` decorator
- remove client side permissions checks since admins can overwrite the default perms
- add `default_member_permissions` attribute
- add `dm_permission` attribute
- add `slash_default_member_permission` decorator	
- add docs for perms and checks

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/models/snek/application_commands.py
+++ b/dis_snek/models/snek/application_commands.py
@@ -196,6 +196,16 @@ class InteractionCommand(BaseCommand):
 
         super().__attrs_post_init__()
 
+    def to_dict(self) -> dict:
+        data = super().to_dict()
+
+        if self.default_member_permissions is not None:
+            data["default_member_permissions"] = str(int(self.default_member_permissions))
+        else:
+            data["default_member_permissions"] = None
+
+        return data
+
     @property
     def resolved_name(self) -> str:
         """A representation of this interaction's name."""
@@ -501,6 +511,8 @@ class SlashCommand(InteractionCommand):
                 group_description=group_description or self.group_description,
                 sub_cmd_name=sub_cmd_name,
                 sub_cmd_description=sub_cmd_description,
+                default_member_permissions=self.default_member_permissions,
+                dm_permission=self.dm_permission,
                 options=options,
                 callback=call,
                 scopes=self.scopes,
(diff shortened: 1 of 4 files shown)


docs/src/Guides/03 Creating Commands.md
```diff
@@ -301,8 +301,114 @@ You are in luck. There are currently four different ways to create interactions,
...
+        MyExtension(bot)
+    ```
+
+    The check will be checked for every command in the extension.
+
+This example only allows the bot owner to execute a command.
+```py
+
+
+```
```
AstreaTSS: I think you forgot to put in the actual example 😅

Kigstn: Woops

AstreaTSS (commented):
Weirdly enough, when using `dm_permission=True`, the slash command (a basic top-level command) doesn't appear in DMs, at least for me. I'm not sure if this is a library or Discord issue, however.

Everything else looks good!

dis_snek/models/snek/application_commands.py
```diff
@@ -922,14 +890,12 @@ def squash_subcommand(subcommands: List) -> Dict:
...
             if not output_data:
-                perms = [s.to_dict() if not isinstance(s, dict) else s for s in subcommand.permissions]
-                perms_filtered = [dict(t) for t in {tuple(d.items()) for d in perms}]
                 output_data = {
                     "name": str(subcommand.name),
                     "description": str(subcommand.description),
                     "options": [],
-                    "permissions": perms_filtered,
-                    "default_permission": subcommand.default_permission,
+                    "default_member_permissions": str(int(subcommand.default_member_permissions)),
```
LordOfPolls: This will raise a `TypeError` for any command that has this set to None. 
image
Replication steps: Load debug scale -> attempt to sync

LordOfPolls (changes requested):
After fixing the highlighted bug, this works fine. Nice work!

LordOfPolls:
> Weirdly enough, when using `dm_permission=True`, the slash command (a basic top-level command) doesn't appear in DMs, at least for me. I'm not sure if this is a library or Discord issue, however.
> 
> Everything else looks good!

Ive tested this - the discord client will not automatically load the new command list/perms for dms without a full-reload, or the cache expiring. The bot and discord are syncing correctly, but the ui wont update unless you wait a while, or hit `ctrl+r`

PR #474, opened 2022-05-11
feat: add `unavailable` bool to `GuildUnavailable`

i0bs:

- [ ] Breaking code change
- [ ] Tests change

## Description
Sometimes, an ["Unavailable Guild" object](https://discord.com/developers/docs/resources/guild#unavailable-guild-object-example-unavailable-guild) returned from the API can also return an `unavailable` boolean denoting whether it's accessible or not. For clarification, this was internally explained that this is a value given for events such as guild termination via. TnS and can represent whether it's actually unavailable by means of the server being deleted, or if the `GUILD_CREATE` event was unable to return data at the time of its dispatch for whatever API issues may be present.

## Changes
- Added an `unavailable` attribute to the `GuildUnavailable` object.

## Checklist

- [x] I've tested my code

### DIFF
--- a/dis_snek/api/events/discord.py
+++ b/dis_snek/api/events/discord.py
@@ -235,6 +235,8 @@ class GuildUnavailable(BaseEvent, GuildEvent):
 
     guild: Optional["Guild"] = field(default=MISSING)
     """The guild, if it was cached"""
+    unavailable: Optional[bool] = field(default=MISSING)
+    """The guild's state, whether accessible temporarily or on the time of dispatch."""
 
 
 @define(kw_only=False)


Wolfhound905:
Please target to naff branch.

LordOfPolls:
I don't believe this field is necessary. The `GuildUnavailble` event is *only* ever triggered when the guild is unavailable. 
Effectively adding this field means there are now 2 places the event is indicating the guild is unavailble, one by the simple fact the event was dispatched, the second by this field. 

It is worth noting that the api event `guild_delete` is resolved into 2 naff/events - `GuildLeft` and `GuildUnavailble`. These events are dispatched according to the unavailble state of the guild.

i0bs:
Are you determining its dispatch by field, event type or both?

I admit that the use case of wanting that field is very small for the TnS purpose, I mostly did this PR simply for schema consistency.

LordOfPolls:
https://github.com/Discord-Snake-Pit/Dis-Snek/blob/naff/naff/api/events/processors/guild_events.py#L59

i0bs:
fair enough. only valid thing i've left to say is schema consistency, but.. meh

LordOfPolls:
I dont think thats necessary, ill leave this pr open, for a few days, to allow other people to give input - in case they disagree, but provisionally im rejecting this.

PR #476, opened 2022-05-11
feat: expose `Application` attrs for App Discovery

i0bs:

- [ ] Breaking code change
- [ ] Tests change

## Description
This simply exposes the `tags` and `custom_install_url` fields which were recently amended to the `Application` object when application discovery was announced. These also may have just been simply overlooked as well.

Side note: I've commented out another attribute missing, `install_params` which needs an object suited for it. I'm willing to implement said object for this PR if one desires

## Changes
- Added `tags` optional attribute for a list of string values
- Added `custom_install_url` optional attribute for a string value

## Checklist
- [x] I've tested my code


i0bs:
It also seems that there's a merge conflict for this file only when switching to `naff` branch too, FYI. No issues on `dev` branch if targeted.

LordOfPolls:
Set install_params as a dict for now. The to-do to create the object stands. Better to at least leave the data accessible than discard it

PR #496, opened 2022-05-19
feat: add asset.as_url

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Adds `asset.as_url` which allows the user to override the extension or size of an asset, rather than using the default `Asset.url` method.

Req by @Astrea49 

## Changes

- Add `asset.as_url` method

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/discord/asset.py
+++ b/naff/models/discord/asset.py
@@ -65,6 +65,9 @@ class Asset:
         """
         if not extension:
             extension = ".gif" if self.animated else ".png"
+        elif not extension.startswith("."):
+            extension = f".{extension}"
+
         return f"{self._url}{extension}?size={size}"
 
     @property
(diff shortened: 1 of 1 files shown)


naff/models/discord/asset.py
```diff
@@ -52,6 +52,21 @@ def url(self) -> str:
...
+        args:
+            extension: The extension to override the assets default with
+            size: The size of asset to return
+
+        returns:
+            A url for this asset with the given parameters
+        """
+        if not extension:
+            extension = ".gif" if self.animated else ".png"
+        return f"{self._url}{extension}?size={size}"
```
AstreaTSS: Suggesting many people are going to do `extension = "jpg"` or the like (at least I did), it's probably worthwhile to make it add the dot before the extension automatically. A couple of lines would need to be changed, but it wouldn't be hard.

AstreaTSS (changes requested):

naff/models/discord/asset.py
LordOfPolls: Resolved in 8cea7bf95b3b4996f00eb919c32378480c02e795 to support both usages

AstreaTSS (approved):
...forgot how simple a solution can be.

PR #498, opened 2022-05-20
fix: DM messages not having channel attrib

### LINKED ISSUE #497: [BUG] `Message.reply` fails in DMs + no channel for DMs
**Describe the bug**
I know, weird wording, but the first part is what people run into 90% of the time, while the second is the *actual* underlying issue.

Anyways, users have recently told us that trying to reply to a message in DMs fails. Furthermore, the error message shows that it's unable to get the DM channel needed in the first place, as it returns `None`.

It was reported that this was working fine in earlier versions, so there may be a regression here. I am unsure if this bug appears in the `master` branch, but it does for sure in `dev`.

**To Reproduce**
```python
After updating I am getting an error upon DM replies. The repro would be doing something like 

```python
@listen()
async def on_message_create(self, event: MessageCreate) -> None:
    if not event.message.guild:
        await event.message.reply("Pong!")
```

**Expected behaviour**
The user to be DMed 'Pong!'

Instead, we get a basic "'NoneType' has no attribute 'send'."

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Resolves #497 

If the DM channel has never been cached, the library will leave the channel attrib blank. The `create_dm` endpoint is very API-light so i see no issue adding this call. Not to mention it would only be called once per session, or when the cache expires. 

## Changes

- Add call to `fetch_dm_channel` when a dm message is recieved but no DM Channel has been cached

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/events/processors/message_events.py
+++ b/naff/api/events/processors/message_events.py
@@ -39,6 +39,11 @@ class MessageEvents(EventMixinTemplate):
             else:
                 await self.cache.fetch_user(to_snowflake(msg._author_id))
 
+        # noinspection PyProtectedMember
+        if not msg._guild_id and not msg.channel:
+            # dm message with no channel cached
+            await self.cache.fetch_dm_channel(msg.author.id)
+
         self.dispatch(events.MessageCreate(msg))
 
     @Processor.define()


LordOfPolls:
Closing this pr. I've decided that it isnt worth the chance that someone could be flagged for API abuse via an exploit or just a lotta dms thanks to the library automatically fetching DM channels. Instead if you need reference to the DM channel in a command or something, call `cache.fetch_dm_channel`

PR #499, opened 2022-05-21
Private channel fix

### LINKED ISSUE #497: [BUG] `Message.reply` fails in DMs + no channel for DMs

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Another potential fix for the forbidden channel issue. This time its a dummy channel object. I appreciate this isnt an ideal solution, but as this dummy channel object is never cached, it won't cause future api calls to be skipped. 

## Changes

- On forbidden, `cache.fetch_channel` will generate a `BaseChannel` of type MISSING with only the ID field populated
- If a user needs a DM channel reference, and its not cached, a dummy DM channel object will be instantiated 

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/smart_cache.py
+++ b/naff/client/smart_cache.py
@@ -442,8 +442,12 @@ class GlobalCache:
         channel_id = to_snowflake(channel_id)
         channel = self.channel_cache.get(channel_id)
         if channel is None:
-            data = await self._client.http.get_channel(channel_id)
-            channel = self.place_channel_data(data)
+            try:
+                data = await self._client.http.get_channel(channel_id)
+                channel = self.place_channel_data(data)
+            except Forbidden:
+                log.warning(f"Forbidden access to channel {channel_id}. Generating fallback channel object")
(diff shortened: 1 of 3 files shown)


LordOfPolls:

PR #500, opened 2022-05-22
fix: Ensure subprocess is killed on linux

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

On Linux the Audio subproccess was not exiting successfully, resulting in zombie processes being left behind. This *should* resolve this, without impacting Windows. 

I also found that the `read_ahead` task was not being cancelled either, this fixes that too.

## Changes

- Set subprocess `stdin` to `DEVNULL`
- Check if child process is runnning in `_read_ahead`
- Add check for child process state in `cleanup`

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/voice/audio.py
+++ b/naff/api/voice/audio.py
@@ -141,7 +141,9 @@ class Audio(BaseAudio):
             f"{self.ffmpeg_args}".split()
         )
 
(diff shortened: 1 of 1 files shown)


LordOfPolls:

PR #501, opened 2022-05-22
fix: Allow audio/args with spaces in them

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

The dumby who made voice originally was lazy and just used `.split` on the command string for subprocess. This meant any audio with spaces in the name would not play. This PR fixes that, and also allows lists to be passed as args without breaking.

This change should not break any existing bots, only enable missing functionality 

## Changes

- Restrucutred ffmpeg command to be a list from the start
- Args are now accepted as list or string
- Args are injected into the command as needed

## Checklist

- [x] I've tested my code


LordOfPolls:

PR #505, opened 2022-05-27
fix: remove unnecessary Activity converter

Wolfhound905:

- [ ] Breaking code change
- [ ] Tests change

## Description

So from what I was able to understand. It looked like we were converting an Activty object into an Activity object and it was breaking it lol. 
image
^ like this, it put all the data in the name.

So removing the converter from the user object seems to work.

## Changes

- Remove the broken converter from User object

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/discord/user.py
+++ b/naff/models/discord/user.py
@@ -119,9 +119,7 @@ class User(BaseUser):
         converter=optional_c(Color),
         metadata=docs("The user's banner color"),
     )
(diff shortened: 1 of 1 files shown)


LordOfPolls:
This will break caching the startup activity data. You need to add an `isinstance` check to prevent the object nesting rather than this

Wolfhound905:
I don't know what you mean.

PR #506, opened 2022-05-27
feat: add orjson as optional & dev dependencies

Wolfhound905:

- [ ] Breaking code change
- [ ] Tests change

## Description

I think that this would be useful for some of the common tools used in committing to the library. Black is a big one, and should be ran before every commit, and pre-commit can help too. I also added orjson as an extra because that is true to the library when you start a bot without orjson.

## Commands

| Goal | Command |
| ------------- | ------------- |
| Install without dev deps | `poetry install --no-dev` |
| Install optional w/o dev deps | `poetry install --no-dev -E orjson` |
| install all | `poetry install  -E orjson` |

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = "aiohttp"
-version = "3.8.0"
+version = "3.8.1"
 description = "Async http client/server framework (asyncio)"
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


zevaryx (approved):
Happy?

Wolfhound905:
> Happy?

I assume he wanted you to review it because not sure he uses the poetry lol

LordOfPolls:
I havent used poetry actively in a while, zev does. made sense to make zev check this

PR #530, opened 2022-06-14
NAFF 1.3.0

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Merges Dev into Master

## Checklist

- [x] I've tested my code


leestarb (approved):
Good

Kigstn:
Shouldn't this be `1.2.1` since no new feats got added? IMO type hinting improvements should be a `fix` / `style` / `refactor` and not a `feat`

LordOfPolls:
As you can see in the above log. I originally opted for 1.2.1. 

However the reason the typing changes got lumped into feat was because it was the closest match in conventional commits. Fix, style and refactor aren't relevant to typing 

Fix being resolving a bug
Style being a change that doesn't affect the meaning of the code (ie formatting)
Refactor being a change that neither adds a feature or fixes a bug 

Refactor is closest but as typing adds a feature, being an IDE feature, it felt most appropriate. Imo if not feat, docs is most fitting, but seeing as this is a code change its null.

Kigstn (approved):
Fair enough. At the end, it doesn't really matter much imo

silasary (approved):

PR #535, opened 2022-06-20
fix: add protected attribs to user.pyi

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
Just adding back some protected attributes to `user.pyi` that the generator I used ignored. While these weren't really used by a typical user, it's still useful to typehint them.

## Changes
Both of these are for `user.pyi`.
- Add `_guild_ids` to `User`. I used `Set` over `set` to keep in line with `user.py`.
- Add `_guild_id` and `_role_ids` to `Member`.

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/discord/user.pyi
+++ b/naff/models/discord/user.pyi
@@ -13,7 +13,7 @@ from naff.models.discord.role import Role
 from naff.models.discord.snowflake import Snowflake_Type
 from naff.models.discord.timestamp import Timestamp
 from naff.models.discord.voice_state import VoiceState
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Set
 
 class _SendDMMixin(SendMixin):
     id: Snowflake_Type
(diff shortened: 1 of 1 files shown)


LordOfPolls:
I disagree with these being added to the pyi files.

Pyi files outline the *public* interface, private attributes are most assuredly not that.

AstreaTSS:
> I disagree with these being added to the pyi files.
> 
> Pyi files outline the _public_ interface, private attributes are most assuredly not that.

Disagree there. `.pyi` files are simply for typehinting - nowhere does it state it *only* outlines the public interface, [at least in the PEP](https://peps.python.org/pep-0484/#stub-files). The closest you can get to something saying that is `mypy` claiming that, but that's also a (kinda) third-party library with its own opinions. Even Python's own [`typeshed`](https://github.com/python/typeshed) uses them to typehint underscored/protected properties.

In this case, I'd argue this is making it so the typehints are actually correct, without regard to caring if they're protected or not. If you wish to put a standard on `.pyi` files to make them *only* reflect the public interface, that's fair enough, even if I disagree with it. But that's a personal choice.

LordOfPolls:
Reasonable. I'll stand corrected

LordOfPolls (approved):

PR #538, opened 2022-06-21
feat: Support multiple clients in the same process

### LINKED ISSUE #290: [FEAT] Allow Multiple Bot Instances In The Same Process
**Is your feature request related to a problem? Please describe.**
When creating multiple bot instances, only one continues to send heartbeats and remain active.

**Describe the solution you'd like**
The best solution would be the ability to create as many instances of Snake, and use login() in a task to have them all connect separately.

**Describe alternatives you've considered**
The alternatives all would require creating separate processes and using pipes or IPC to communicate between instances which is extreme pain.

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

This PR potentially fixes #290. It achieves this by making all objects that `callback` subclass a `CallbackObject` which allows the binding to be passed during initialisation. The client now also stores a copy of `CallbackObject` rather than modifying the original, resolving the problem where clients were overwriting eachothers callbacks. 

A known issue with this solution is Tasks. Tasks dont hook themselves into the client, and are instead just methods that get called periodically by asyncio. Due to this, they cannot be bound in the same way as other callbacks, such as commands and listeners. The only solutions I can come up with for this, would be breaking, which the PR goes out of its way to attempt to avoid. As such they have not been modified 

## Changes

- Impliment `CallbackObject`
- Make Listner, Command, PrefixedCommand subclass `CallbackObject`
- Use `CallbackObject` binding rather than `wrap_partial`

## Checklist

- [x] I've tested my code


LordOfPolls:
Regarding Tasks, if you choose to use tasks **and** mutliple clients in one process, you will need to move your tasks to an extension to ensure correct binding

Kigstn (approved):
Seems to work, but looks like magic

PR #539, opened 2022-06-26
feat: Add auto-shard support

LordOfPolls:
- [ ] Breaking code change
- [ ] Tests change

## Description
Adds a new `AutoShardedClient` which will automatically create and manage all requested shards with minimal syntax changes, and importantly respecting discord rate limits in sharding and identification. This client allows all shards to share the same cache. 

This PR also refines the guild caching step in the startup sequence, preventing erroneous "Not All Guilds Cached" warnings. 

## Changes
- Create `AutoShardedClient`
- Create `ShardConnect` and `ShardDisconnect` events
- Change logging messages to include shard where relevant 
- Add `_shard_ready` event to `ConnectionState`
- Add `get_gateway_bot` method to httpclient
- Refactor `Client.intents` into a attribute rather than property
- Add `client`/`AutoShardedClient“.get_guild_web socket` to allow getting the appropriate web socket for a given guild
- Refactor `Client._on_websocket_ready` to avoid false warnings
- Fix logger name

## Checklist
- [x] I've tested my code


Wolfhound905:
How many shard would you suggest the max be for this auto sharding?

LordOfPolls:
Technically as many as you need. However in my opinion this form of sharding is only suitable for 2-10 shards, after that more processes would be needed.

naff/api/events/internal.py
```diff
@@ -100,6 +102,15 @@ class Disconnect(BaseEvent):
     """The bot has just disconnected."""
 
 
+class ShardConnect(Connect):
+    """A shard just connected to the discord Gateway."""
```
Bluenix2, line 107: This won't have slots and instead get a `__dict__`. Did you mean for that to happen?

Shouldn't you add `@define()` at the top?

naff/api/gateway/state.py
```diff
@@ -91,13 +96,17 @@ async def stop(self) -> None:
...
             async with GatewayClient(self, (self.shard_id, self.client.total_shards)) as self.gateway:
                 try:
                     await self.gateway.run()
                 finally:
-                    self.client.dispatch(events.Disconnect())
+                    self._shard_ready.clear()
+                    if self.client.total_shards == 1:
+                        self.client.dispatch(events.Disconnect())
+                    else:
+                        self.client.dispatch(events.ShardDisconnect())
```
Bluenix2: I would expect both to be dispatched
```suggestion
                    if self.client.total_shards == 1:
                        self.client.dispatch(events.Disconnect())

                    self.client.dispatch(events.ShardDisconnect())
```

naff/api/gateway/state.py
```diff
@@ -29,6 +29,8 @@ class ConnectionState:
     """The event intents in use"""
     shard_id: int
     """The shard ID of this state"""
+    _shard_ready: asyncio.Event = field(default=None)
+    """indicates this state is now ready"""
```
Bluenix2:
```suggestion
    """Indicates that this state is now ready"""
```

naff/client/auto_shard_client.py
```diff
@@ -0,0 +1,230 @@
+import asyncio
+import logging
+import time
+from collections import defaultdict
+from typing import (
+    TYPE_CHECKING,
+)
```
Bluenix2: This won't expend if you remove the trailing comma.

naff/client/auto_shard_client.py
```diff
@@ -0,0 +1,230 @@
...
+        else:
+            self.auto_sharding = False
+
+        super().__init__(*args, **kwargs)
+
+        self._connection_state = None
+
+        self._connection_states: list[ConnectionState] = []
+
+        self.max_start_concurrency: int = 1
```
Bluenix2, line 47: Shouldn't this be a parameter?

naff/api/gateway/state.py
```diff
@@ -29,6 +29,8 @@ class ConnectionState:
     """The event intents in use"""
     shard_id: int
     """The shard ID of this state"""
+    _shard_ready: asyncio.Event = field(default=None)
```
Bluenix2, line 32: I can't find where you set this event?

naff/client/auto_shard_client.py
```diff
@@ -0,0 +1,230 @@
...
+                log.debug(f"Starting {shard.shard_id}")
+                start = time.perf_counter()
+                tasks.append(asyncio.create_task(shard.start()))
+
+                if self.max_start_concurrency == 1:
+                    # connection ratelimiting when discord has asked for one connection concurrently
+                    # we could wait for `state._shard_ready`, but by waiting on the raw websocket ready
+                    # we can start our subsequent shards sooner - speeding up startup
+                    while not shard.gateway:
+                        await asyncio.sleep(0.1)
```
Bluenix2: Don't you have an event just for this purpose?

naff/client/auto_shard_client.py
```diff
@@ -0,0 +1,230 @@
...
+                    # noinspection PyProtectedMember
+                    await shard.gateway._ready.wait()
+                    await asyncio.sleep(5.1 - (time.perf_counter() - start))
```
Bluenix2, line 196: What is this arbitrary amount of sleep?

naff/api/events/internal.py
LordOfPolls: Thats very true, I did miss the define, thanks. 

Fixed in 5224b343c1348d2758e535a0d3400218042554f2

naff/api/gateway/state.py
LordOfPolls: I disagree; disconnect means the entire bot has disconnected, when shard_disconnect is dispatched, the bot could still be connected. I should add a condition to handle all shards being disconnected, however.

naff/client/auto_shard_client.py
LordOfPolls: Looks like pycharm weirdness. Fixed in 165df325d3e44a7aca517c5d38a114ff163e0bad

naff/client/auto_shard_client.py
LordOfPolls: It was intentionally not. It could be, but protecting this attribute didnt seem necessary, as the only time you could affect anything with it, would be if you manually login, then start the gateway. That would be an advanced usecase, in which case I'm inclined to allow the modification.

naff/api/gateway/state.py
LordOfPolls: https://github.com/NAFTeam/NAFF/blob/auto-shard/naff/client/auto_shard_client.py#L136

naff/client/auto_shard_client.py
LordOfPolls: Sadly not

naff/client/auto_shard_client.py
LordOfPolls: Its not arbitrary

Discord requires 5 seconds between subsiquent batches of connections. Through testing i found this drifted by about 0.1s, likely down to asyncio imprecision and discord's servers. 5.1 seconds allowed the bot to successfully connect without being hit by discord invalidating the session.

naff/client/auto_shard_client.py
LordOfPolls: Further looking at this code, i dont like the polling loop. im replacing it

Kigstn (changes requested):
Missing some documentation about basic usage IMO: 
1) what this does
2) when it should be used
3) when it should no longer be used (in favor of multiple workers)

Kigstn (approved):

PR #541, opened 2022-06-27
feat: Add a SendableContext protocol.

silasary:

- [ ] Breaking code change
- [x] Tests change

## Description
Adds a protocol that is essentially "Context + SendMixin"
This is useful for type-hinting anything that works in either Message or Interaction contexts.

## Checklist

- [x] I've tested my code

### DIFF
--- a/.github/workflows/pytest-pr.yml
+++ b/.github/workflows/pytest-pr.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install pytest
         run: |
           pip install -e ${{ matrix.extras }}
-          pip install .tests
+          pip install .[tests]
       - name: Run Tests
         env:
           BOT_TOKEN: ${{ secrets.BOT_TOKEN }}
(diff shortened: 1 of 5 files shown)


silasary:
Let me know if this feels like the wrong file to put it in

LordOfPolls:
No qualms with this outside of the pre-commit fails

.github/workflows/pytest-pr.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Install pytest
         run: |
           pip install -e ${{ matrix.extras }}
-          pip install pytest pytest-recording pytest-asyncio pytest-cov
```
Kigstn, line 58: Depends on how strict polls is, but technically this does not belong in this PR

LordOfPolls: Not feeling strict at all

silasary: I included it because I didn't want to add typeguard to both files ;)

LordOfPolls (approved):

PR #544, opened 2022-06-28
feat: Implement CursedIntEnum

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

This PR introduces the `CursedIntEnum` i threatened to add a short while ago. This modifies `IntEnum` to allow for any value to be passed without throwing an exception. The purpose of this is to prevent bots losing functionality when discord adds or changes something.

## Changes

- Move `ChannelTypes` cursed converter into its own class using the `_missing_` magic method
- Make all `intEnum` classes inherit from `cursedIntEnum`
- Remove `ChannelTypes` calls for `converter`

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/discord/enums.py
+++ b/naff/models/discord/enums.py
@@ -2,7 +2,7 @@ import logging
 from enum import Enum, EnumMeta, IntEnum, IntFlag, _decompose
 from functools import reduce
 from operator import or_
-from typing import Iterator, Tuple
+from typing import Iterator, Tuple, TypeVar
 
 from naff.client.const import logger_name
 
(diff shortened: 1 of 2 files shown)


naff/models/discord/enums.py
```diff
@@ -77,7 +77,21 @@ def __iter__(self) -> Iterator:
         yield from _decompose(self.__class__, self)[0]
 
 
-class WebSocketOPCodes(IntEnum):
+SELF = TypeVar("SELF")
+
+
+class CursedIntEnum(IntEnum):
+    @classmethod
+    def _missing_(cls, value) -> SELF:
```
Bluenix2: This typing with `SELF` is not correct.
```suggestion
    def _missing_(cls: Type[SELF], value) -> SELF:
```

Kigstn (approved):

PR #547, opened 2022-06-30
feat: simple logging

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Introduces a simple logging config that allows an end-user to quickly get logging to console without understanding the logging library.

## Changes

- Add `basic_logging` client arg to activate `logging.basic_config`
- Add `logging_level` client arg to configure the logging level

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/client.py
+++ b/naff/client/client.py
@@ -154,6 +154,8 @@ class Client(
 
         debug_scope: Snowflake_Type: Force all application commands to be registered within this scope
         asyncio_debug: bool: Enable asyncio debug features
+        basic_logging: bool: Utilise basic logging to output library data to console
+        logging_level: int: The level of logging to use for basic_logging
 
     Optionally, you can configure the caches here, by specifying the name of the cache, followed by a dict-style object to use.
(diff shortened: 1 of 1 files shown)


Kigstn:
Better yet, how about a client arg where users can pass their own logger which we would use. Makes it especially nice for people like me who like using `rich.Logger`

zevaryx:
I agree with @Kigstn 

I currently have to override all of the error handling logic in order to get meaningful error messages while using rich logging. While not the end of the world, having a solution that lets you pass a logger (or logger class) in would be nice

LordOfPolls:
That's not easily possible. All the files with logging would need to somehow reference your custom logic class, which will cause nightmares of circular imports.

The alternative is to just have it affect logging in the client or error handler, which seems redundant as you'd now have 2 loggers

Note: never used this rich logging thing, if I'm missunderstanding its usage lmk

LordOfPolls:
Iirc I believe that was pred before and it was rejected by vote, also

silasary (approved):

PR #548, opened 2022-06-30
Feat: Logging alternative to #547

Kigstn:

- [ ] Breaking code change
- [ ] Tests change

## Description

This is an alternative solution to #547, attempting to solve the logging problem.

This PR allows users to pass their own logger at client initialisation. The passed logger is saved in both the client and naff.constants.

This has several benefits;
1) users don't have to do nasty stuff at the very start of their code to overwrite the naff logger before naff does anything with it
2) custom loggers, like `rich.logging` or file handlers can be easily used
3) users have more direct and easier access to the logger via both `client.logger` and `naff.consts.logger`, instead of having to `logging.getLogger(logger_name)` every time they want to use it
4) naff devs have an easier time as well, since they can import the logger from anywhere as well

## Changes

- a logger obj is now defined in `naff.consts` which is overwritten by the logger the user passes to `Client`

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/client.py
+++ b/naff/client/client.py
@@ -398,7 +398,7 @@ class Client(
 
     def _sanity_check(self) -> None:
         """Checks for possible and common errors in the bot's configuration."""
-        self.logger.debug("Running client sanity checks...")
+        logger.debug("Running client sanity checks...")
         contexts = {
             self.interaction_context: InteractionContext,
             self.prefixed_context: PrefixedContext,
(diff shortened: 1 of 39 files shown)


naff/client/client.py
```diff
@@ -193,8 +195,12 @@ def __init__(
         **kwargs,
     ) -> None:
 
-        # Configuration
+        # Set Up logger and overwrite the constant
+        self.logger = logger
+        """The logger NAFF should use"""
+        constants.logger = logger
```
silasary, line 208: How does this interact with multiple clients in the same process?
Can I define a separate logger for each, and will it actually work as expected?

(I assume not, but I feel the need to raise the question)

Kigstn: Only one logger is supported, it won't work. Theoretically it should be possible by passing the client obj around to literally every function, but that would be a lot of work to make that change for a very niche feature

LordOfPolls: Pending resolution of this comment before i can merge

silasary: I'm fine with the answer being "this isn't supported". But a simple mention of that fact in the docs would be appreciated

LordOfPolls:
On-Hold pending discussion and conflict resolution

naff/client/client.py
Kigstn: Good point, I will put that in the docstring

Kigstn:
Ready for review

LordOfPolls (approved):
It hurts me calling it `logger`

Kigstn:
> It hurts me calling it `logger`

It hurts me calling it `log`

PR #559, opened 2022-07-11
fix: check min/max value in sync comparison

### LINKED ISSUE #557: [BUG] Command Syncing
### NAFF Version

any

### Describe the Bug

When using slash command options max and mins and you change just `max_value` or others the command does not sync

### Expected Results

The command should sync with the new min_value

### Minimal Reproducible Code

```python
@naff.slash_command("int")
@naff.slash_option(
    "int",
    "test",
    naff.OptionTypes.INTEGER,
    min_value=0,
    max_value=15, # Just change this value and command wont sync
)
async def test(ctx, string):
    await ctx.send(string)
    return "test"
```

### Traceback

_No response_

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

### Additional Information

_No response_

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Resolves #557 by chceking the min and max attributes for options. 

## Changes

- Adds `min_value` and `max_value` to the attributes that are compared in sync detection 

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/naff/application_commands.py
+++ b/naff/models/naff/application_commands.py
@@ -1054,6 +1054,8 @@ def _compare_options(local_opt_list: dict, remote_opt_list: dict) -> bool:
         "name_localized": ("name_localizations", {}),
         "description_localized": ("description_localizations", {}),
         "choices": ("choices", []),
+        "max_value": ("max_value", None),
+        "min_value": ("min_value", None),
     }
 
     if local_opt_list != remote_opt_list:


silasary:
Is the test failing because of rate limits?

silasary (approved):

LordOfPolls:
> Is the test failing because of rate limits?


No, its a race condition in the test-bot. 

Its editing the user, then checking to see if the change was successful. Sometimes it checks the value before discord sends the data.

silasary:
Ah, we should definitely make the tests more resilient to that 😛

Issue #403, opened 2022-03-17
[FEAT] Port Sentry integration from taipan to main library


silasary:
This is relatively easy, I'm just waiting for Voice to get merged before I do it.

Wolfhound905:
@silasary hey is there any update on this?

artem30801:
In my opinion, all external integrations should not be in this repo (even in ext), but instead be placed in separate repos under the same org, to be optional dependencies

silasary:
I've been thinking about that, and I'm very torn.  Having stuff in external repos reduces discoverability, and makes the features a lot less likely to be used.

`naff_audio` must be separate for messy reasons, but for most extensions, I'd rather them be relatively painless and selling points for the library, rather than hidden, generally unknown 2nd-party plugins.

Where possible, I'd much rather have them shipped in `ext`, and have any additional dependencies locked behind an extra, rather than "This [really cool feature](https://github.com/artem30801/nafftrack) exists, but you'd never know it unless you happened to be looking at Discord the one day a year someone brings it up"

compil3:
I would definitely use if it were shipped with the library, I already do but having less things to manually configure that are quality of life improvements, in my eyes are well worth it.

PR #573, opened 2022-07-28
feat: Convert `client.on_error` into a proper event

silasary:
- [ ] Breaking code change
- [ ] Tests change

## Description
This makes the error handler a dispatchable event, allowing Extensions listen to all events (prereq for #403)

## Changes
Adds new Internal Event `naff.api.events.internal.Error`
Changes all calls to `client.on_error` to dispatch that event instead
Adds a non-breaking listener `client._on_event` that calls the existing method.
Also fixes implausible bug when `on_command_error` is called without a valid bot token.

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/events/internal.py
+++ b/naff/api/events/internal.py
@@ -168,8 +168,8 @@ class Select(Component):
 class Error(BaseEvent):
     """Dispatched when the library encounters an error."""
 
-    source: str
-    error: Exception
+    source: str = field(metadata=docs("The source of the error"))
+    error: Exception = field(metadata=docs("The error that was encountered"))
     args: tuple[Any] = field(factory=tuple)
     kwargs: dict[str, Any] = field(factory=dict)
-    ctx: Optional["Context"] = field(default=None)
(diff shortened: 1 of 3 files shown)


naff/api/events/internal.py
```diff
@@ -161,3 +162,14 @@ class Button(Component):
...
+
+@define(kw_only=False)
+class Error(BaseEvent):
+    """Dispatched when the library encounters an error."""
+
+    source: str
+    error: Exception
+    args: tuple[Any] = field(factory=tuple)
+    kwargs: dict[str, Any] = field(factory=dict)
+    ctx: Optional["Context"] = field(default=None)
```
LordOfPolls: Please can you docstring these attributes?

LordOfPolls (changes requested):

naff/api/events/internal.py
silasary: I'm honestly not sure how to document args and kwargs.

PR #577, opened 2022-07-31
Feat: Intent warning for events and prefixed commands

Kigstn:

- [ ] Breaking code change
- [ ] Tests change

## Description

This warns users if they use a gateway event / prefixed commands and do not have the required intents enabled. If users read their console output, this should make debugging easier for them.

Lists of intents and events taken from https://discord.com/developers/docs/topics/gateway#list-of-intents

## Changes

- warn when event is missing required intent
- warn when prefixed commands are used but the message content intent is missing

## Checklist

- [x] I've tested my code


LordOfPolls:
I'd suggest using references to the event objects rather than strings. That way if we were to refactor they'd be caught by most modern IDEs

Kigstn:
> I'd suggest using references to the event objects rather than strings. That way if we were to refactor they'd be caught by most modern IDEs

Sure, good call

AstreaTSS:
Too lazy to do a proper review, but it should be noted that prefixed commands will work fine even without the message content intent *unless* they change the prefixes used to outside of their defaults (you can always mention the bot). Probably worth checking for that.

naff/client/client.py
```diff
@@ -33,7 +33,55 @@
 
 import naff.api.events as events
 import naff.client.const as constants
-from naff.api.events import RawGatewayEvent, MessageCreate
+from naff.api.events import (
```
silasary: We are already importing the events namespace on line 34, let's just use that, and refer to events as `events.Whatever`, the same way we do when dispatching them.

silasary (changes requested):

silasary (approved):

naff/client/client.py
```diff
@@ -33,9 +33,7 @@
 
 import naff.api.events as events
 import naff.client.const as constants
-from naff.api.events import RawGatewayEvent, MessageCreate
-from naff.api.events import processors
-from naff.api.events.internal import Component, BaseEvent
+from naff.api.events import MessageCreate, RawGatewayEvent, processors, Component, BaseEvent
```
silasary, line 36: Bonus points if you refactor all existing references to these to also use `events.`, but I'm happy to merge without that

PR #583, opened 2022-08-02
fix: detect no arg converters even if its a class

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
Previously, the prefixed command parser was not detecting `NoArgumentConverter` and its subclasses if it was left in its raw class form (at least, it wasn't when using my WIP hybrid commands. `CMD_*` should have been messed up with it too, but I haven't heard complains of that... whatever). This fixes that by adding extra conditions to check if the annotation is a subclass, instead of just an instance of a class.

It also uses `TypeGuard`, which is cool.

## Changes

- Move out `NoArgumentConverter` checking to its own function.
- Add in an extra check to see if the annotation was possibly a class and a subclass of `NoArgumentConverter`.

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/naff/prefixed_commands.py
+++ b/naff/models/naff/prefixed_commands.py
@@ -3,7 +3,7 @@ import inspect
 import typing
 from collections import deque
 from types import NoneType, UnionType
-from typing import Optional, Any, Callable, Annotated, Literal, Union, TYPE_CHECKING, Type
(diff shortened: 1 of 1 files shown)


AstreaTSS:
Closing this as there are multiple problems with `NoArgumentConverter`, isn't that wonderful?

PR #585, opened 2022-08-02
feat: add hybrid commands

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
This PR adds hybrid commands, which allow using a slash command as a prefixed command too. Efforts have been taken to make sure the resulting prefixed command is as close to the slash command as possible.

This PR also, mostly by consequence:
- Adds in `HybridContext`. An argument has been added to `Client` to allow setting a custom subclass of `HybridContext`, too.
- Allows using custom signatures for prefixed commands. This is a very low level thing and should only be manipulated by those who know what they're doing.
- Changes prefixed commands to base if it should use a kwarg for a parameter or not based on its parameter type instead of blindly guessing.
- Adds `kind` to `PrefixedCommandParameter`.
- Adds docs for hybrid commands.

Due to the many, *many* ways slash commands can be declared, there may be some bugs that I missed (though I did test a number of combinations and options).

## Changes

See above. Usually, I would type something out, but I feel like this would be a repeat of the description if I bothered.

## Checklist

- [x] I've tested my code


silasary:
Can you go into [tests/test_protocols.py](https://github.com/NAFTeam/NAFF/blob/dev/tests/test_protocols.py) and a check that Hybrid contexts are SendableContexts?

AstreaTSS:
Done!

naff/models/naff/extension.py
```diff
@@ -86,6 +86,8 @@ def __new__(cls, bot: "Client", *args, **kwargs) -> "Extension":
                         bot.add_modal_callback(val)
                     elif isinstance(val, naff.ComponentCommand):
                         bot.add_component_callback(val)
+                    elif isinstance(val, naff.HybridCommand):
+                        bot.add_hybrid_command(val)
```
LordOfPolls, line 90: You forgot to handle commands being *un*loaded from extensions

AstreaTSS: Fixed! Though the fix is pretty complex 😅.

LordOfPolls (approved):

LordOfPolls:
Due to the scope of this PR it should not be merged until we have at least a second trust-worthy reviewer

That said if it hits 14 days since my review without a second, it'll be merged

AstreaTSS:
Found a little oversight - `min/max_length` wasn't being respected. That's been fixed with the latest two commits.

PR #588, opened 2022-08-07
feat: add client.owner_ids

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Adds support for declaring owner ids in code. This allows users to add other users as owners, without creating a team. 

## Changes

- Add `client.owner_ids` attribute
- Update `is_owner` check to respect `client.owner_ids`
- Add API-True owner to `owner_ids` on startup

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/naff/checks.py
+++ b/naff/models/naff/checks.py
@@ -62,7 +62,7 @@ def has_id(user_id: int) -> TYPE_CHECK_FUNCTION:
 
 def is_owner() -> TYPE_CHECK_FUNCTION:
     """
-    Is the author the owner of the bot.
+    Checks if the author is the owner of the bot. This respects the `client.owner_ids` list.
 
     Args:
         coro: the function to check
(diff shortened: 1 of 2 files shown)


naff/models/naff/checks.py
```diff
@@ -72,7 +72,7 @@ def is_owner() -> TYPE_CHECK_FUNCTION:
     async def check(ctx: Context) -> bool:
         if ctx.bot.app.team:
             return ctx.bot.app.team.is_in_team(ctx.author.id)
```
silasary: This currently ignores owner_ids if the bot is in a team, which feels unintuitive.  We should instead add the entire team to owner_ids, and then always check that.

silasary (changes requested):

naff/models/naff/checks.py
LordOfPolls: Resolved in 7bf04825333e1d4b0a86d3c57d528bd1c6fed023

silasary (approved):

PR #601, opened 2022-08-13
feat: implement REST chunking

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

This PR changes the default chunking routine to use the REST api instead of gateway commands. 

The main benifit of this is that we can chunk guilds far faster. Through my testing I have found using REST chunking takes approximately 30% less time than traditional gateway chunking. The main bottleneck remains actually creating the member objects. 
The reasoning why this is faster is the fact that the gateway has far more intense ratelimits than the REST api. This means the client is able to make substantionally more requests in the same interval (while still respecting all ratelimts). On top of this, when gateway-chunking, you must wait for each chunk to arrive at a relatively slow rate.  

## Changes

- Create `guild.http_chunk`
- Create `guild.chunk` which calls `guild.http_chunk`
- Seperate logic from `guild.chunk_guild` into `guild.gateway_chunk`
- Depreciate `guild.chunk_guild`
- Convert `guild_create` processor to use new `guild.chunk` methoid

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/events/processors/guild_events.py
+++ b/naff/api/events/processors/guild_events.py
@@ -41,7 +41,7 @@ class GuildEvents(EventMixinTemplate):
 
         if self.fetch_members:  # noqa
             # delays events until chunking has completed
-            await guild.chunk_guild(presences=True)
+            await guild.chunk()
 
         self.dispatch(events.GuildJoin(guild))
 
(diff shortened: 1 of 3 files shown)


silasary:
Is Gateway Chunking the only way to get presences?
If so, we probably want to set that to True by default, and add that note to a docstring somewhere.

LordOfPolls:
Yes gateway chunking is the only way to get presence data during startup. Ill set the default to True

silasary (approved):

PR #613, opened 2022-08-22
feat: add dict postprocessing in sync comparison

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Discord loves not sending data, we know that. This pr adds a post_processing system to option comparison, which currently is the problem-child. 

Post processing, in effect, adds any missing keys that discord decided weren't worth sending. The system has been built to be scalable, simply add the field name that needs processing, and a lambda function to inact the change. 

I'm marking this as a discord bug as this is discord deciding not to send data, causing us to have to handle their issues. 

## Changes

- Add post_processing dict to compare_options
- Call lambda function when comparing options

## Checklist

- [x] I've tested my code


LordOfPolls:
This PR is being closed as its changes are being integrated into another

PR #630, opened 2022-09-09
feat: support component only messages

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Adds support for sending messages with only components
https://github.com/discord/discord-api-docs/pull/5435

## Changes

- Add components to the minimum required content in a message send

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/mixins/send.py
+++ b/naff/client/mixins/send.py
@@ -1,6 +1,5 @@
 from typing import TYPE_CHECKING, Iterable, List, Optional, Union
 
-import naff.client.errors as errors
 import naff.models as models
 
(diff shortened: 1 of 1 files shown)


Kigstn (approved):

Kigstn:
Depending on how legible the error is that discord returns when a message is fully empty, it may be worth removing the check completely

LordOfPolls:
There arent really any **reasonable** ways to send messages that would raise an exception. Except `send()`

I'm removing the check - however ive noticed a bug with error parsing thanks to discord, making a seperate pr for that

PR #632, opened 2022-09-09
feat: restore asyncio debug in debug ext

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

This pr restores the debug ext behaviour of setting asyncio to its DEBUG mode and enabling tracemaloc. This functionality was removed when we migrated to not using the loop variable in accordance with asyncio. This adds it in a compliant manor.

## Changes

- Set asyncio to debug in an async start task
- Start tracemaloc on ext load
- Log the above at warning level 

## Checklist

- [x] I've tested my code


LordOfPolls:
retargetting to 2.x

PR #639, opened 2022-09-10
fix: fix usage of hybrid cmds with no localized names

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change

## Description
Seems like Polls fixed Discord localization syncing by making `to_locale_dict` return `None` if the there's no localization. This screwed over assumptions hybrid commands made, so this PR ensures that there will always be a dict to work with.

## Changes
- Make every usage of `to_locale_dict` use `to_locale_dict() or {}` for hybrid commands. This requires some ugly parenthesis, but otherwise works fine.

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/client.py
+++ b/naff/client/client.py
@@ -1128,7 +1128,7 @@ class Client(
             prefixed_base = self.prefixed_commands.get(str(command.name))
             if not prefixed_base:
                 prefixed_base = _base_subcommand_generator(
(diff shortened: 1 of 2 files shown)


LordOfPolls:
Polls did this because Discord was stupid - apologies this broke hybrid. Merging

PR #641, opened 2022-09-11
feat💥: Switch error and completion function to proper events

Kigstn:

- [x] Breaking code change
- [ ] Tests change

## Description

See the nice bullet list below.

There still is the issue that neither the `ComponentError` nor the `ModalError` event will be called if `ctx.wait_for_component` or `ctx.wait_for_modal` are used. This was a problem before this however and needs some thinking how best to fix.

## Changes

- add `CommandCompletion` event
- add `ComponentCompletion` event
- add `AutocompleteCompletion` event 
- add `ModalCompletion` event
- add `CommandError` event
- add `ComponentError` event
- add `AutocompleteError` event 
- add `ModalError` event
- allow events to be overwritable by setting `delete_if_overwritten`. This means they will be deleted if the user registeres their own listener for that event
- (💥) remove `ModalResponse` event in favor of the new `ModalCompletion` event
- (💥) remove `on_command` calls in favor of the new `CommandCompletion` event
- (💥) remove `on_component` calls in favor of the new `ComponentCompletion` event
- (💥) remove `on_autocomplete` calls in favor of the new `AutocompleteCompletion` event
- (💥) remove `on_command_error` calls in favor of the new `CommandError` event
- (💥) remove `on_component_error` calls in favor of the new `ComponentError` event
- (💥) remove `on_autocomplete_error` calls in favor of the new `AutocompleteError` event
- (💥) remove `on_error` calls in favor of the `Error` event

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/events/__init__.py
+++ b/naff/api/events/__init__.py
@@ -1,6 +1,7 @@
 from . import processors
 from .discord import *
 from .internal import *
+from .base import *
 
 __all__ = (
     "RawGatewayEvent",
--- a/naff/api/events/base.py
+++ b/naff/api/events/base.py
@@ -1,0 +1,83 @@
+import re
+from typing import TYPE_CHECKING, Callable, Coroutine
+
+import naff.models as models
+from naff.client.const import MISSING
+from naff.client.utils.attr_utils import define, field, docs
+from naff.models.discord.snowflake import to_snowflake
+
+if TYPE_CHECKING:
+    from naff import Client
+    from naff.models.discord.snowflake import Snowflake_Type
+    from naff.models.discord.guild import Guild
+
+__all__ = ("BaseEvent", "GuildEvent", "RawGatewayEvent")
+
+_event_reg = re.compile("(?<!^)(?=[A-Z])")
+
+
+@define(slots=False)
+class BaseEvent:
+    """A base event that all other events inherit from."""
+
+    override_name: str = field(kw_only=True, default=None)
+    """Custom name of the event to be used when dispatching."""
+    bot: "Client" = field(kw_only=True, default=MISSING)
+    """The client instance that dispatched this event."""
+
+    @property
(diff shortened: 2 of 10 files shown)


silasary (changes requested):
Please add listeners to `naff.ext.sentry` for the new error events.

Kigstn:
> Please add listeners to `naff.ext.sentry` for the new error events.

Which ones? Functionality is the same here as on pypi, since the new error events all dispatch `Error` by default which sentry handles. On pypi, the `Error` event is also only called if the user has not overwritten it

silasary:
Ah, yep.  Never mind then.  I just misread the code when doing my initial review.

silasary (approved):

silasary:
One thing I'd like to clarify though:

If I override a listener function that has `delete_if_overwritten`, I can still expect it to work?  Because from reading it, it seems closer to `delete_if_subscribed` than `delete_if_overwritten`

Kigstn:
> One thing I'd like to clarify though:
> 
> If I override a listener function that has `delete_if_overwritten`, I can still expect it to work?  Because from reading it, it seems closer to `delete_if_subscribed` than `delete_if_overwritten`

Yes. I'm not really happy with the name, happy to take suggestions. If the flag is set, the listener will be deleted if a listener without the flag is added. 

The event will still trigger, but only the listener without the flag will be called.

silasary:
Let's figure out a better name for it before we merge then.

LordOfPolls:
`delete_if_overriden` would be an apt name, failing that `delete_if_subscribed` would be a valid name 
`delete_if_overwritten` would be a misnomer

Kigstn:
I'll change it to `delete_if_overridden` for now then

LordOfPolls (approved):

PR #643, opened 2022-09-12
fix💥: make naff always use the correct logger

Kigstn:

- [x] Breaking code change
- [ ] Tests change

## Description

Naff was not always using the correct logger. This was due to the logger being imported before the imported variable was changed on Client initialisation. 

`const.logger` has been converted to a function which ensures the fetched object is the set one.

Technically breaking if anyone imported and used the logger

## Changes

- convert `const.logger` to a function

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/const.py
+++ b/naff/client/const.py
@@ -86,9 +86,15 @@ __repo_url__ = "https://github.com/Discord-Snake-Pit/NAFF"
 __py_version__ = f"{_ver_info[0]}.{_ver_info[1]}"
 __api_version__ = 10
 logger_name = "naff"
-logger = logging.getLogger(logger_name)
+_logger = logging.getLogger(logger_name)
 default_locale = "english_us"
 kwarg_spam = False
+
+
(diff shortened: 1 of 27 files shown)


naff/client/const.py
```diff
@@ -86,10 +86,16 @@
 __py_version__ = f"{_ver_info[0]}.{_ver_info[1]}"
 __api_version__ = 10
 logger_name = "naff"
-logger = logging.getLogger(logger_name)
+_logger = logging.getLogger(logger_name)
 default_locale = "english_us"
 kwarg_spam = False
 
+
+def logger() -> logging.Logger:
```
silasary, line 94: If you make this an `@property`, you don't need to change everything to `logger()`, and it won't be a breaking change.

Kigstn: Properties only work properly in instantiated classes tho, which this is not. Or am I missing something here?
image

PR #646, opened 2022-09-15
feat 💥: Forums rewrite

LordOfPolls:

- [x] Breaking code change
- [ ] Tests change

## Description

Rewrites / ammends the forum implimentation to be more user friendly, and inline with the spec

## Changes

- Update create post endpoint from private-beta code to spec
- Add `NewThreadCreate` event that only dispatches for brand new threads
- Add `list_posts` method to forums
- Add `fetch_post` method to forums
- Add `get_post` method to forums
- Add `get_tag` method to forums
- Update forum `create_post` to handle string, int, and object available tags

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/models/discord/channel.py
+++ b/naff/models/discord/channel.py
@@ -2360,6 +2360,8 @@ class GuildForum(GuildChannel):
                     if not tag:
                         continue
                     tag = tag.id
+                elif isinstance(tag, dict):
+                    tag = tag["id"]
                 processed.append(tag)
 
             applied_tags = processed
--- a/naff/models/discord/channel.py
+++ b/naff/models/discord/channel.py
@@ -2425,7 +2425,7 @@ class GuildForum(GuildChannel):
 
         return next((tag for tag in self.available_tags if predicate(tag)), None)
 
-    async def create_tag(self, name: str, emoji: Union["models.PartialEmoji", dict, str]) -> "ThreadTag":
+    async def create_tag(self, name: str, emoji: Union["models.PartialEmoji", dict, str, None] = None) -> "ThreadTag":
         """
         Create a tag for this forum.
 
@@ -2440,15 +2440,20 @@ class GuildForum(GuildChannel):
             The created tag object.
 
         """
-        if isinstance(emoji, str):
-            emoji = PartialEmoji.from_str(emoji)
-        elif isinstance(emoji, dict):
-            emoji = PartialEmoji.from_dict(emoji)
+        payload = {"channel_id": self.id, "name": name}
 
-        if emoji.id:
-            data = await self._client.http.create_tag(self.id, name, emoji_id=emoji.id)
-        else:
-            data = await self._client.http.create_tag(self.id, name, emoji_name=emoji.name)
+        if emoji:
+            if isinstance(emoji, str):
+                emoji = PartialEmoji.from_str(emoji)
+            elif isinstance(emoji, dict):
+                emoji = PartialEmoji.from_dict(emoji)
+
+            if emoji.id:
+                payload["emoji_id"] = emoji.id
+            else:
+                payload["emoji_name"] = emoji.name
+
+        data = await self._client.http.create_tag(**payload)
 
         channel_data = self._client.cache.place_channel_data(data)
         return [tag for tag in channel_data.available_tags if tag.name == name][0]
--- a/naff/models/discord/channel.py
+++ b/naff/models/discord/channel.py
@@ -1977,11 +1977,11 @@ class GuildForumPost(GuildPublicThread):
         Args:
             name: 1-100 character channel name
             archived: whether the thread is archived
-            applied_tags: list of tags to apply to a forum post (!!! This is for forum threads only)
+            applied_tags: list of tags to apply
             auto_archive_duration: duration in minutes to automatically archive the thread after recent activity, can be set to: 60, 1440, 4320, 10080
             locked: whether the thread is locked; when a thread is locked, only users with MANAGE_THREADS can unarchive it
             rate_limit_per_user: amount of seconds a user has to wait before sending another message (0-21600)
-            flags: channel flags for forum threads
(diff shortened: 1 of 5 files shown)


Kigstn (changes requested):
Gonna test this over the next days. But for now support for pinning posts is missing

naff/models/discord/channel.py
```diff
@@ -2280,6 +2290,65 @@ async def create_post(
         )
         return self._client.cache.place_channel_data(data)
 
+    async def list_posts(self) -> List["GuildPublicThread"]:
```
Kigstn: Should this not be called ``fetch_posts`, since it does a request?

naff/models/discord/channel.py
```diff
@@ -2280,6 +2290,65 @@ async def create_post(
...
+
+        Returns:
+            A list of GuildPublicThread objects representing the posts.
+        """
+        data = await self._client.http.list_active_threads(self._guild_id)
+        threads = [self._client.cache.place_channel_data(post_data) for post_data in data["threads"]]
+
+        return [thread for thread in threads if thread.parent_id == self.id]
+
+    async def fetch_post(self, id: "Snowflake_Type") -> "GuildPublicThread":
```
Kigstn: While it's technically a thread, it may be worth thinking about creating a `ForumPost` model, since threads do not have many things forums have (pins, tags, initial post, etc...) 
While this would be different than the API does it, IMO it would help with ease of use since users would, for example, not try to use tags on normal threads

naff/models/discord/channel.py
```diff
@@ -2218,7 +2217,7 @@ async def create_post(
         self,
         name: str,
         content: str | None,
-        applied_tags: Optional[List[Union["Snowflake_Type", "ThreadTag"]]] = MISSING,
+        applied_tags: Optional[List[Union["Snowflake_Type", "ThreadTag", str]]] = MISSING,
```
Kigstn, line 2321: Should allow a single tag to be passed without putting it in a list

Kigstn (changes requested):

naff/models/discord/channel.py
LordOfPolls: Hard disagree, the attribute is `applied_tags`, plural. Unlike components, id rather enforce passing the correct data structure for tags - plus its **really** not a big ask to wrap your payload in a list before using it

naff/models/discord/channel.py
```diff
@@ -1904,8 +1910,54 @@ async def edit(
...
+        self,
+        name: Absent[str] = MISSING,
+        archived: Absent[bool] = MISSING,
+        auto_archive_duration: Absent[AutoArchiveDuration] = MISSING,
+        locked: Absent[bool] = MISSING,
+        rate_limit_per_user: Absent[int] = MISSING,
+        flags: Absent[Union[int, ChannelFlags]] = MISSING,
+        reason: Absent[str] = MISSING,
+        **kwargs,
+    ) -> "GuildPublicThread":
```
Kigstn, line 1951: Should be `GuildForumPost`

Kigstn (changes requested):
`create_tag()` currently requires an emoji, this needs to be optional

naff/models/discord/channel.py
```diff
@@ -2280,6 +2357,74 @@ async def create_post(
         )
         return self._client.cache.place_channel_data(data)
 
+    async def fetch_posts(self) -> List["GuildForumPost"]:
```
Kigstn, line 2396: `fetch_posts` only returns active posts, although the name suggests all posts will be fetched. Fetching archived posts should either be made an argument, or default behavior IMO.

Whatever happens a method to fetch archived posts is missing

Kigstn (changes requested):

naff/models/discord/channel.py
LordOfPolls: Archived threads were intentionally left out from this method (and the get equiv) because using it would be very expensive. Im open to adding a tertiary method (async iterator like other objects get) but im very opposed to including it in this method

naff/models/discord/channel.py
LordOfPolls: No it shouldnt, look at the class name

LordOfPolls: Im assuming you meant to comment on line 1973, which is wrong

5cdc66240e8c5c9d17fd96ade89ba95098cdd829

naff/models/discord/channel.py
Kigstn: Sure, that works!

naff/models/discord/channel.py
```diff
@@ -2254,10 +2320,21 @@ async def create_post(
...
-            A GuildPublicThread object representing the created post.
+            A GuildForumPost object representing the created post.
         """
         if applied_tags != MISSING:
-            applied_tags = [str(tag.id) if isinstance(tag, ThreadTag) else str(tag) for tag in applied_tags]
+            processed = []
+            for tag in applied_tags:
+                if isinstance(tag, ThreadTag):
+                    tag = tag.id
+                if isinstance(tag, (str, int)):
```
Kigstn: This needs to be `elif`. Currently fails if a list of thread tags is passed

naff/models/discord/channel.py
```diff
@@ -1926,11 +1977,11 @@ async def edit(
         Args:
             name: 1-100 character channel name
             archived: whether the thread is archived
-            applied_tags: list of tags to apply to a forum post (!!! This is for forum threads only)
+            applied_tags: list of tags to apply
```
Kigstn, line 2014: After a post is edited with new tags, the cache does not get updated with them
image
image

LordOfPolls: Interesting. Would have thought gateway would have taken care of that. I'll look into it

Kigstn (approved):

PR #647, opened 2022-09-16
fix💥: Make naff always use the correct logger (v2)

Kigstn:

- [x] Breaking code change
- [ ] Tests change

## Description

Naff was not always using the correct logger. This was due to the logger being imported before the imported variable was changed on Client initialisation. 

`const.logger` has been converted to a function which ensures the fetched object is the set one.

Technically breaking if anyone imported and used the logger

## Changes

- convert `const.logger` to a function

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/gateway/gateway.py
+++ b/naff/api/gateway/gateway.py
@@ -7,7 +7,7 @@ from types import TracebackType
 from typing import TypeVar, TYPE_CHECKING
 
 from naff.api import events
-from naff.client.const import logger, MISSING, __api_version__
+from naff.client.const import MISSING, __api_version__
 from naff.client.utils.input_utils import OverriddenJson
 from naff.client.utils.serializer import dict_filter_none
 from naff.models.discord.enums import Status
@@ -176,31 +176,31 @@ class GatewayClient(WebsocketClient):
         match op:
 
             case OPCODE.HEARTBEAT:
-                logger().debug("Received heartbeat request from gateway")
+                self.logger.debug("Received heartbeat request from gateway")
                 return await self.send_heartbeat()
 
             case OPCODE.HEARTBEAT_ACK:
                 self.latency.append(time.perf_counter() - self._last_heartbeat)
 
                 if self._last_heartbeat != 0 and self.latency[-1] >= 15:
-                    logger().warning(
+                    self.logger.warning(
                         f"High Latency! shard ID {self.shard[0]} heartbeat took {self.latency[-1]:.1f}s to be acknowledged!"
                     )
                 else:
-                    logger().debug(f"❤ Heartbeat acknowledged after {self.latency[-1]:.5f} seconds")
+                    self.logger.debug(f"❤ Heartbeat acknowledged after {self.latency[-1]:.5f} seconds")
 
                 return self._acknowledged.set()
 
             case OPCODE.RECONNECT:
-                logger().debug("Gateway requested reconnect. Reconnecting...")
+                self.logger.debug("Gateway requested reconnect. Reconnecting...")
                 return await self.reconnect(resume=True, url=self.ws_resume_url)
 
             case OPCODE.INVALIDATE_SESSION:
-                logger().warning("Gateway has invalidated session! Reconnecting...")
+                self.logger.warning("Gateway has invalidated session! Reconnecting...")
                 return await self.reconnect()
 
             case _:
-                return logger().debug(f"Unhandled OPCODE: {op} = {OPCODE(op).name}")
+                return self.logger.debug(f"Unhandled OPCODE: {op} = {OPCODE(op).name}")
 
     async def dispatch_event(self, data, seq, event) -> None:
         match event:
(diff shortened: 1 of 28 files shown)


LordOfPolls:
This does not appear to work:

All existing projects i open encounter this error:
```python
Traceback (most recent call last):
  File "C:\Users\LordOfPolls\AppData\Local\JetBrains\Toolbox\apps\PyCharm-P\ch-0\222.3739.56\plugins\python\helpers\pydev\pydevd.py", line 1496, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:\Users\LordOfPolls\AppData\Local\JetBrains\Toolbox\apps\PyCharm-P\ch-0\222.3739.56\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/LordOfPolls/Documents/Projects/NAFF/_main_test.py", line 12, in <module>
    bot = naff.Client(
  File "C:\Users\LordOfPolls\Documents\Projects\NAFF\naff\client\client.py", line 409, in __init__
    self._sanity_check()
  File "C:\Users\LordOfPolls\Documents\Projects\NAFF\naff\client\client.py", line 503, in _sanity_check
    self.logger.debug("Running client sanity checks...")
AttributeError: 'function' object has no attribute 'debug'
```

Kigstn:
> This does not appear to work:
> 
> All existing projects i open encounter this error:
> 
> ```python
> Traceback (most recent call last):
>   File "C:\Users\LordOfPolls\AppData\Local\JetBrains\Toolbox\apps\PyCharm-P\ch-0\222.3739.56\plugins\python\helpers\pydev\pydevd.py", line 1496, in _exec
>     pydev_imports.execfile(file, globals, locals)  # execute the script
>   File "C:\Users\LordOfPolls\AppData\Local\JetBrains\Toolbox\apps\PyCharm-P\ch-0\222.3739.56\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
>     exec(compile(contents+"\n", file, 'exec'), glob, loc)
>   File "C:/Users/LordOfPolls/Documents/Projects/NAFF/_main_test.py", line 12, in <module>
>     bot = naff.Client(
>   File "C:\Users\LordOfPolls\Documents\Projects\NAFF\naff\client\client.py", line 409, in __init__
>     self._sanity_check()
>   File "C:\Users\LordOfPolls\Documents\Projects\NAFF\naff\client\client.py", line 503, in _sanity_check
>     self.logger.debug("Running client sanity checks...")
> AttributeError: 'function' object has no attribute 'debug'
> ```

Apparently I only tested this with a logger supplied. There was one parenthesis missing

naff/api/gateway/gateway.py
```diff
@@ -176,31 +176,31 @@ async def dispatch_opcode(self, data, op: OPCODE) -> None:
...
                 return await self.reconnect(resume=True, url=self.ws_resume_url)
 
             case OPCODE.INVALIDATE_SESSION:
-                logger.warning("Gateway has invalidated session! Reconnecting...")
+                logger().warning("Gateway has invalidated session! Reconnecting...")
                 return await self.reconnect()
 
             case _:
-                return logger.debug(f"Unhandled OPCODE: {op} = {OPCODE(op).name}")
+                return logger().debug(f"Unhandled OPCODE: {op} = {OPCODE(op).name}")
```
LordOfPolls: Gunna be honest; syntax like this is horrific. I am strongly opposed to having to use this. Not to mention the fact this is an incredibly niche issue

Kigstn: I agree that the notation is not optimal. Fancier solution would be passing the client obj to every class, which would also resolve this. 
Open for better suggestions. 

However, I disagree with the niche issue part. The passed logger should always be respected.

LordOfPolls: Yes, the passed logger should be respected. but the act of passing a logger is niche in itself - this has been a common point throughout this saga 

As of right now, i dont have a better suggestion, however im very opossed to merging syntax like this, regardless of the fact it may work

LordOfPolls: Hell, why not just stick to the conventional paradigm of a `get_logger` method in const, which then every class that needs a logger calls to have a `self.log` attrib. This would eliminate this ugly syntax, be pythonic, and make sense to a reader

LordOfPolls: Further, you could expand this so all `ClientObject`s do this in their base form

Kigstn: While this would fix logging calls from all classes, there are multiple instances where the logger is used and the client is not easily accessible (without changing a lot of function calls) - something I'd like to avoid doing because the chance of breaking something seems high. 

`logger()` was already used whenever the `Client` is available, I have added it to all remaining places where that's not the case and wasn't be too big of a change.

LordOfPolls: I don't think you've understood what I was suggesting. I'm at work, I'll make a proof of concept later

LordOfPolls: So 
const.py
```python
def get_logger():
    ...
   return logger
```

mixins
```python
class DictSerializationMixin:
    def __init__():
        self.logger = get_logger()
```

ergo in any class subclassing from that
```python
class someclass(DictSerializationMixin:
    ...
    def some_func(self):
        self.log.debug("some log message")
```

Then in any class you need a logger that doesnt inherit, just run the same code to define a logger attribute. This isnt that disimilar from your solution, buts its so much less... ick

Kigstn: Ahhh, thanks for clearing that up!

naff/client/mixins/serialization.py
```diff
@@ -30,7 +33,7 @@ def _get_init_keys(cls) -> frozenset:
     def _filter_kwargs(cls, kwargs_dict: dict, keys: frozenset) -> dict:
         if const.kwarg_spam:
             unused = {k: v for k, v in kwargs_dict.items() if k not in keys}
-            const.logger.debug(f"Unused kwargs: {cls.__name__}: {unused}")  # for debug
+            const.get_logger.debug(f"Unused kwargs: {cls.__name__}: {unused}")  # for debug
```
LordOfPolls: This looks like a mistake

Kigstn: Good catch!

LordOfPolls (approved):

PR #650, opened 2022-09-18
fix💥: handle errors when parsing http exceptions

LordOfPolls:
breaking in case someone was relying on these exceptions

- [x] Breaking code change
- [ ] Tests change

## Description

This prevents httpexception raising an exception when parsing the error payload, instead falling back on the provided text field. 

## Checklist

- [x] I've tested my code


LordOfPolls:
The check fails are unrelated, but ironically they prove this pr works

PR #652, opened 2022-09-22
feat: Add jurigged hot reloading to debug extension

zevaryx:

- [ ] Breaking code change
- [ ] Tests change

## Description

[jurigged](https://github.com/breuleux/jurigged) is a library that can be used to patch Python code in memory, allowing for users to edit their bots while it is running and have the changes sync in real time with no downtime. This PR adds to the DebugExtension to allow jurigged to run and modify the bot as it is running, including re-syncing any commands that need to be resynced.

## Changes

- Add jurigged to DebugExtension

## Checklist

- [x] I've tested my code


zevaryx:
[testbot.zip](https://github.com/NAFTeam/NAFF/files/9621779/testbot.zip)

Here is the code I used to test this. In `testbot/exts/test.py`, you can add/remove commands, arguments, etc. You can also add other files with extensions in them to have them automatically load and sync

pyproject.toml
```diff
@@ -13,6 +13,7 @@ attrs = "^21.4.0"
 mypy = ">0.930"
 discord-typings = "^0.5.1"
 tomli = "^2.0.1"
+jurigged = "^0.5.3"
```
silasary: This also needs to be added to requirements.txt

silasary (changes requested):

pyproject.toml
zevaryx: Done

silasary (approved):

LordOfPolls:
Purely because of how much of a selling point this is, and how complex it is under the hood, can i suggest you make a guide for its usage?

zevaryx:
> Purely because of how much of a selling point this is, and how complex it is under the hood, can i suggest you make a guide for its usage?

I'll get a PR open (or add a commit) to update the docs for the Debug cog to explain exactly what it is

silasary:
Is there any way to disable it? 

Depending on what I'm working on, I can foresee times when I would want the rest of DebugExtension, but juryrigged would cause invalid runtime states

zevaryx:
> Is there any way to disable it? 
> 
> Depending on what I'm working on, I can foresee times when I would want the rest of DebugExtension, but juryrigged would cause invalid runtime states 

Not by default. I could add a flag to disable it. I didn't add that since the rest of the debug extensions didn't have that, either

silasary:
The rest of the debug extensions don't have any potential downsides of having them enabled ;)

LordOfPolls:
Any movement re the above comments @zevaryx?

zevaryx:
I had asked about where to put the documentation in the Discord and never got a response. I'm probably going to create a subsection of the docs explaining all of the debug extensiona

silasary:
The last discussion we had on discord was about moving it out of DebugExension, and making it a separate extension entirely

zevaryx:
Yes! I have those changes made locally, but was waiting for the conversations on documentation. Been busy at work these last few weeks, but I'll get the changes finished and pushed this week

LordOfPolls (approved):

naff/ext/debug_extension/utils.py
```diff
@@ -164,3 +165,30 @@ def make_table(rows: list[list[Any]], labels: Optional[list[Any]] = None, center
         lines.append(_make_data_line(column_widths, row, data_left, data_middle, data_right, align))
     lines.append(_make_solid_line(column_widths, "╰", "┴", "╯"))
     return "\n".join(lines)
+
+
+def get_all_commands(module: ModuleType) -> Dict[str, Callable]:
```
silasary: Why is this in this file?

silasary (changes requested):

naff/ext/debug_extension/utils.py
zevaryx: I couldn't think of a better place to put it at the time. It's a useful utility regardless of being used in jurigged or not, and when I added it it was next to the jurigged extension.

This could be moved to `naff.utils` if you think that's a better spot

silasary: Yeah, that feels better.

zevaryx: Done in latest commit, it's now in `naff.client.utils.misc_utils` with other search-like functions

silasary (approved):

docs/src/API Reference/ext/index.md
```diff
@@ -5,6 +5,9 @@ These files contain useful features that help you develop a bot
 - [Debug Extension](debug_ext)
     - An extension preloaded with a load of debugging utilities to help you find and fix bugs
 
+- [Jurigged](jurigged)
```
silasary, line 8: This name means nothing to people who don't already know the library.  I'd rather call it "Live Editing" or "Hot Reloading" or something similar.

zevaryx: I think "Live Patching" is a better name

zevaryx: However, for the API reference, keeping the `jurigged` name should be fine; I'll update the other parts of the docs to reference it as Live Patching

PR #667, opened 2022-10-15
feat 💥: differentiate an unavailable guild coming back online

LordOfPolls:

- [x] Breaking code change
- [ ] Tests change

## Description

This allows NAFF to differentiate between GUILDCREATE events for new joins and unavailable guilds returning. 
Breaking as this changes when `GuildJoin` is dispatched... but im pretty sure this behaviour isn’t desirable anyways

## Changes

- Check if the guild is already cached, if so, it was almost certainly an unavailable guild 
- Adds a new event: `GuildAvailable` which is dispatched when a guild comes back online, instead of `GuildJoin`

## Checklist

- [x] I've tested my code


Kigstn:
Not sure I like this, since behaviour would depend on user configs / cache state. I just fear that this will lead to confusion because the same guilds will trigger different events depending on the state. Additionally, the name of the events does imo not reflect that the logic behind it is basically a glorified guess and not a fact.

LordOfPolls:
The issue is; discord gives no way for a library to determine if a `GUILDCREATE` event is a new guild or not. This is because library's/apps are expected to keep a cache of their guilds to handle that. While yes NAFF gives you the option to not do that, it should be anticipated it will cause issues. 

With that in mind, even though a user could break this by disabling the guild cache, id argue any bot with a disabled guild cache shouldn''t be expecting accurate guild state events. 

Its the same principle that you shouldn't be checking for changes in roles if you disable the role cache. Or message edits without a message cache. 

Regarding the event naming, I think you're stretching there. The names are accurate regardless of how the state is determined,

PR #679, opened 2022-10-17
fix: resolve global ratelimit being exceeded

### LINKED ISSUE #677: Bot has exceeded global ratelimit, locking REST API for 0.002 seconds
Sentry Issue: [INQUIRY-7](https://sentry.io/organizations/inquiry/issues/3676005886/?referrer=github_integration)

```
Bot has exceeded global ratelimit, locking REST API for 0.002 seconds
```

Related: https://github.com/LordOfPolls/Inquiry/issues/8

LordOfPolls:

- [ ] Breaking code change
- [ ] Tests change

## Description

Resolves #677 

Prevents the HTTP client from exceeding the global rate limit. The previous implementation could be exceeded if you fired a lot of requests in quick succession... defeating the point of the limit. 

For the base implementation of this, credit should be given to @Bluenix2 in this thread https://github.com/discord/discord-api-docs/issues/5144#issuecomment-1179471952. Their implementation mostly works, but as per their own comment; they can still hit global rate limits. The most likely cause of this would be the request landing in different global buckets due to exec time. This can be resolved simply by changing the maximum allowed requests to a number below 50. 49 works, however, this implementation uses 45 to be more conservative. 

Through testing, I was unable to hit a 429 Global rate limit. However, as a further safeguard, this also overrides the default `ClientSession.connector` with a `TCPConnector` with a hard limit of `global_lock.max_requests` (45) active connections.  

Test Code:
```python
@listen()
    async def on_ready(self):
        log.info("Bot is ready")

        await asyncio.gather(*[self.http.get_current_bot_information() for i in range(75)])
```

## Changes

- Utilise new rate limit system to global limits
- Overrides connector with one with a 45 active connections limit

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/api/http/http_client.py
+++ b/naff/api/http/http_client.py
@@ -1,10 +1,12 @@
 """This file handles the interaction with discords http endpoints."""
 import asyncio
+import time
 from typing import Any, cast
 from urllib.parse import quote as _uriquote
 from weakref import WeakValueDictionary
 
 import aiohttp
+import discord_typings
 from aiohttp import BaseConnector, ClientSession, ClientWebSocketResponse, FormData
 from multidict import CIMultiDictProxy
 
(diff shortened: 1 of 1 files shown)


Kigstn (approved):

switchupcb:
As the creator of that issue, I will state that the [disgo](https://github.com/switchupcb/disgo) implementation works without limiting the number of requests per second. You will also find that limiting the amount of requests per second doesn't do anything to solve the issue, because requests can land at Discord at any time. Such that this implementation will minimize issues but not eliminate them entirely.

The following Disgo implementation resolves this issue entirely.

- https://github.com/switchupcb/disgo/blob/v10/_contribution/concepts/REQUESTS.md#what-is-a-rate-limit
- https://github.com/switchupcb/disgo/blob/v10/wrapper/ratelimit.go
- https://github.com/switchupcb/disgo/blob/v10/wrapper/ratelimiter.go
- https://github.com/switchupcb/disgo/blob/v10/wrapper/request.go

This test consistently passes at 50 requests per second while attempting to send 101: https://github.com/switchupcb/disgo/blob/v10/wrapper/tests/integration/ratelimit_test.go

PR #687, opened 2022-10-23
feat: Add Python 3.11 support

zevaryx:

- [ ] Breaking code change
- [ ] Tests change

## Description

This changes a few dependencies to allow for Python 3.11 support

### Note
~~Python 3.11.0rc2 is out, with full 3.11 releasing in a few days. This adds preliminary support, but is far from complete. This does **not** break anything on 3.10, but currently causes issues with commands not working properly on 3.11~~

Python 3.11.0 is fully released, and the breaking changes are fixed with the release. The branch is working as intended, and local tests pass with flying colors

## Changes

- Change required Python version to `>=3.10` from `^3.10`
- Update `aiohttp` to 3.8.3
- Copy `enum._decompose` from the Python 3.10.8 source code into `naff/models/discord/enums.py`

## Checklist

- [x] I've tested my code


silasary:
`enum._decompose` is still needed in 3.10, and tests are failing without it

zevaryx:
Hmmmm, I didn't see that locally. I'm going to have to find an alternative to using `enum._decompose` since it's gone

zevaryx:
Looks like another issue is that `super()` was changed to be `super(type, obj)`. From running `pytest` locally:
image

zevaryx:
The above issue was resolved by updating Python 3.11, so whatever the issue was was resolved. The code currently works and is pending Python 3.11.0 release and updating tests to check 3.11 as well

zevaryx:
image

Python 3.11.0 is released, and this branch is passing. Once the docker images are released, we should be able to integrate multi-version testing into the actions to make sure we don't break either version

silasary (approved):
I am happy with this

LordOfPolls (approved):

PR #700, opened 2022-10-28
feat: Text Utils

AlbertUnruh:

- [ ] Breaking code change
- [ ] Tests change

## Description
Adds a function to check whether something was mentioned somewhere.

## Changes

- add `naff.client.utils.text_utils`

## Checklist

- [x] I've tested my code

### DIFF
--- a/naff/client/const.py
+++ b/naff/client/const.py
@@ -8,6 +8,7 @@ Attributes:
     logger_name str: The name of NAFFs default logger. Invalid if a custom logger is passed to `Client` to replace the default logger.
     logger logging.Logger: The logger used throughout NAFF. If a custom logger is passed to `Client`, this obj is replaced with the new logger.
     kwarg_spam bool: Should ``unused kwargs`` be logged.
+    tag_as_mention bool: Should utils interpret ``BaseUser.tag`` as a mention of it.
 
     ACTION_ROW_MAX_ITEMS int: The maximum number of items in an action row.
     SELECTS_MAX_OPTIONS int: The maximum number of options a select may have.
@@ -48,6 +49,7 @@ __all__ = (
     "get_logger",
     "logger_name",
     "kwarg_spam",
+    "tag_as_mention",
     "DISCORD_EPOCH",
     "ACTION_ROW_MAX_ITEMS",
     "SELECTS_MAX_OPTIONS",
@@ -96,6 +98,7 @@ def get_logger() -> logging.Logger:
 
 default_locale = "english_us"
 kwarg_spam = False
+tag_as_mention = True
 
 DISCORD_EPOCH = 1420070400000
 
--- a/naff/client/const.py
+++ b/naff/client/const.py
@@ -8,7 +8,6 @@ Attributes:
     logger_name str: The name of NAFFs default logger. Invalid if a custom logger is passed to `Client` to replace the default logger.
     logger logging.Logger: The logger used throughout NAFF. If a custom logger is passed to `Client`, this obj is replaced with the new logger.
     kwarg_spam bool: Should ``unused kwargs`` be logged.
-    tag_as_mention bool: Should utils interpret ``BaseUser.tag`` as a mention of it.
 
     ACTION_ROW_MAX_ITEMS int: The maximum number of items in an action row.
     SELECTS_MAX_OPTIONS int: The maximum number of options a select may have.
@@ -49,7 +48,6 @@ __all__ = (
     "get_logger",
     "logger_name",
     "kwarg_spam",
-    "tag_as_mention",
     "DISCORD_EPOCH",
     "ACTION_ROW_MAX_ITEMS",
     "SELECTS_MAX_OPTIONS",
@@ -98,7 +96,6 @@ def get_logger() -> logging.Logger:
 
 default_locale = "english_us"
 kwarg_spam = False
-tag_as_mention = True
 
 DISCORD_EPOCH = 1420070400000
 
--- a/naff/client/utils/text_utils.py
+++ b/naff/client/utils/text_utils.py
@@ -1,4 +1,5 @@
 import re
+import naff.client.const as const
 import naff.models as models
 
 __all__ = ("mentions",)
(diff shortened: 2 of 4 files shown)


naff/models/discord/message.py
```diff
@@ -380,6 +381,15 @@ def get_referenced_message(self) -> Optional["Message"]:
             return None
         return self._client.cache.get_message(self._channel_id, self._referenced_message_id)
 
+    def does_mention(self, query: "str | re.Pattern[str] | models.BaseUser | models.BaseChannel | models.Role") -> bool:
```
AlbertUnruh: *I'm open for other names for the method

silasary: Let's call it `contains_mention`

naff/client/utils/text_utils.py
```diff
@@ -0,0 +1,20 @@
...
+def mentions(text: str, query: "str | re.Pattern[str] | models.BaseUser | models.BaseChannel | models.Role") -> bool:
+    """Checks whether a query is present in a text."""
+    # don't use match/case witch type(query) since subclasses aren't checked
+    if isinstance(query, str):
+        return query in text
+    elif isinstance(query, re.Pattern):
+        return query.match(text) is not None
+    elif isinstance(query, models.BaseUser):
+        # mentions with <@!ID> aren't detected without the replacement
+        return (query.mention in text.replace("@!", "@")) or (query.username in text)
```
LordOfPolls: Why are you checking if the username is in the text? `LordOfPolls` isn't a mention `<@174918559539920897>` is

Further, if you removed that superfluous check you could squash this and the below if statements into one

AlbertUnruh: I don't want to check for (Discord) mentions, but for the presence itself (if it makes sense).

naff/models/discord/message.py
AlbertUnruh: I'll change it as soon I'm on my computer

naff/client/utils/text_utils.py
LordOfPolls: That seems counter-intuitive to the feature's name, nor how most people would assume it works. 

Assuming we did go that route, however, using `display_name` would be better surely.

naff/client/utils/text_utils.py
```diff
@@ -0,0 +1,20 @@
+import re
+import naff.models as models
+
+__all__ = ("mentions",)
+
+
+def mentions(text: str, query: "str | re.Pattern[str] | models.BaseUser | models.BaseChannel | models.Role") -> bool:
+    """Checks whether a query is present in a text."""
+    # don't use match/case witch type(query) since subclasses aren't checked
```
LordOfPolls: You need to add the parameters to the docstring

naff/client/utils/text_utils.py
AlbertUnruh: *at least I would check for `BaseUser.tag` -> `LordOfPolls#1010`

naff/client/utils/text_utils.py
AlbertUnruh: I'll add it later *(I'm out for the day, should have time in the evening (UTC+2))*

naff/client/utils/text_utils.py
AlbertUnruh: > That seems counter-intuitive to the feature's name, nor how most people would assume it works. 
> 
> Assuming we did go that route, however, using `display_name` would be better surely. 

Any other suggestions for the features name?

LordOfPolls: I would suggest removing the username detection and keeping the name as it is. ;)

silasary: Username detection really feels prone to false positives.  I definitely wouldn't want it by default.

naff/client/utils/text_utils.py
AlbertUnruh: added with [728bb03](https://github.com/NAFTeam/NAFF/pull/700/commits/728bb038c0bec1c8d7f0a680965238a9089d489e)

naff/client/utils/text_utils.py
AlbertUnruh: like so?
https://github.com/NAFTeam/NAFF/blob/da514aa1f91f422a55224fd7ba0da322b5a90df9/naff/client/utils/text_utils.py#L23-L25

naff/client/const.py
```diff
@@ -96,6 +98,7 @@ def get_logger() -> logging.Logger:
 
 default_locale = "english_us"
 kwarg_spam = False
+tag_as_mention = True
```
silasary: Why is this in const?
It should:
A. Be a parameter on contains_mention
B. Be False by default.

AlbertUnruh: Well... I was tired yesterday

I'll change it later

AlbertUnruh: changed with [03e2f64](https://github.com/NAFTeam/NAFF/pull/700/commits/03e2f64f2de76eca3749fa40cc8dad285ed7789b)

naff/client/utils/text_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    query: "str | re.Pattern[str] | models.BaseUser | models.BaseChannel | models.Role",
+    *,
+    tag_as_mention: bool = False,
+) -> bool:
+    """Checks whether a query is present in a text.
+
+    Args:
+        text: The text to search in
+        query: The query to search for
+        tag_as_mention: Should ``BaseUser.tag`` be checked *(only if query is an instance of BaseUser)*
```
LordOfPolls:
```suggestion
        tag_as_mention: Should `BaseUser.tag` be checked *(only if query is an instance of BaseUser)*
```

naff/models/discord/message.py
```diff
@@ -380,6 +381,24 @@ def get_referenced_message(self) -> Optional["Message"]:
...
+        query: "str | re.Pattern[str] | models.BaseUser | models.BaseChannel | models.Role",
+        *,
+        tag_as_mention: bool = False,
+    ) -> bool:
+        """
+        Check whether the message contains the query or not.
+
+        Args:
+            query: The query to search for
+            tag_as_mention: Should ``BaseUser.tag`` be checked *(only if query is an instance of BaseUser)*
```
LordOfPolls:
```suggestion
            tag_as_mention: Should `BaseUser.tag` be checked *(only if query is an instance of BaseUser)*
```

naff/client/utils/text_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Returns:
+        Whether the query could be found in the text
+    """
+    # don't use match/case witch type(query) since subclasses aren't checked
```
LordOfPolls: Redundant 
```suggestion
```

LordOfPolls (changes requested):
Inconsequential bits, otherwise LGTM

silasary (approved):

PR #730, opened 2022-11-24
perf: adjust parts of message processing

AstreaTSS:

- [ ] Breaking code change
- [ ] Tests change
- [ ] CI change
- [ ] Other: [Replace with a description]

## Description
This PR touches on the `message_create` processor, removing the (hopefully) unneeded `author` patch and adjusting the rest of the code to match.

However, more importantly, this adjusts `_dispatch_prefixed_commands` to use the `raw_message_create` event instead of `message_create`, as there are several cases that can be used to filter out actually processing a message before we serialize it. More specifically, we can filter out common cases like not having any prefixed commands or `content` being empty. While I haven't tested the speed difference, I'm willing to gauge that it is a decent speedup in most cases.

Sadly, we still need to serialize the message relatively early due to `generate_prefixes` using the message as an argument.

## Changes

- For `_on_raw_message_create`: remove the author check and move some of the other checks out of it, since they can still be needed independently of it.
- For `_dispatch_prefixed_commands`:
  - Make it use `raw_message_create` instead of `message_create` and try to detect for a few common cases to not run a prefixed command (the user being a bot, there being no content in the message, having no prefixed commands in the first place) before serializing.
  - Make the rest of the code not use nested if statements, using early returns instead to clean it up.

## Test Scenario(s)

I used a test bot running code from my (Realms Playerlist Bot)[https://github.com/Astrea49/RealmsPlayerlistBot] to make sure everything still functioned how it should, since that has a couple of debug prefixed commands. Also, the core logic of actually processing the content and all hasn't been touched, so it would pretty weird if it *did* break something.

## Checklist

- [x] I've added docstrings to everything I've touched
- [ ] I've ensured my code works on `Python 3.11.x`
- [x] I've tested my changes
- [ ] I've added tests for my code - if applicable
- [ ] I've updated the documentation - if applicable

### DIFF
--- a/naff/client/client.py
+++ b/naff/client/client.py
@@ -1752,90 +1752,114 @@ class Client(
         else:
             raise NotImplementedError(f"Unknown Interaction Received: {interaction_data['type']}")
 
-    @Listener.create("message_create", is_default_listener=True)
-    async def _dispatch_prefixed_commands(self, event: MessageCreate) -> None:
+    @Listener.create("raw_message_create")
+    async def _dispatch_prefixed_commands(self, event: RawGatewayEvent) -> None:
         """Determine if a prefixed command is being triggered, and dispatch it."""
-        message = event.message
-
-        if not message.content:
+        # don't waste time processing this if there are no prefixed commands
+        if not self.prefixed_commands:
             return
 
-        if not message.author.bot:
-            prefixes: str | Iterable[str] = await self.generate_prefixes(self, message)
+        data = event.data
 
-            if isinstance(prefixes, str) or prefixes == MENTION_PREFIX:
-                # its easier to treat everything as if it may be an iterable
-                # rather than building a special case for this
-                prefixes = (prefixes,)  # type: ignore
+        # many bots will not have the message content intent, and so will not have content
+        # for most messages. since there's nothing for prefixed commands to work off of,
+        # we might as well not waste time
+        if not data.get("content"):
+            return
 
-            prefix_used = None
+        # webhooks and users labeled with the bot property are bots, and should be ignored
+        if data.get("webhook_id") or data["author"].get("bot", False):
+            return
 
-            for prefix in prefixes:
-                if prefix == MENTION_PREFIX:
-                    if mention := self._mention_reg.search(message.content):  # type: ignore
-                        prefix = mention.group()
+        # now, we've done the basic filtering out, but everything from here on out relies
+        # on a proper message object, so we do the same thing as the message create
+        # processor
+        message = self.cache.place_message_data(event.data)
+        if not message._guild_id and event.data.get("guild_id"):
+            message._guild_id = event.data["guild_id"]
+
+        if message._guild_id and not message.guild:
+            await self.cache.fetch_guild(message._guild_id)
+
+        if not message.channel:
+            await self.cache.fetch_channel(to_snowflake(message._channel_id))
+
+        # here starts the actual prefixed command parsing part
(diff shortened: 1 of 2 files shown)


naff/client/client.py
```diff
@@ -1752,90 +1752,114 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:
         else:
             raise NotImplementedError(f"Unknown Interaction Received: {interaction_data['type']}")
 
-    @Listener.create("message_create", is_default_listener=True)
-    async def _dispatch_prefixed_commands(self, event: MessageCreate) -> None:
+    @Listener.create("raw_message_create")
```
silasary: This is still a default listener

silasary (changes requested):

naff/client/client.py
AstreaTSS: Fixed!

silasary (approved):

naff/client/client.py
```diff
@@ -1752,90 +1752,114 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:
...
+            return
 
-            if isinstance(prefixes, str) or prefixes == MENTION_PREFIX:
-                # its easier to treat everything as if it may be an iterable
-                # rather than building a special case for this
-                prefixes = (prefixes,)  # type: ignore
+        # now, we've done the basic filtering out, but everything from here on out relies
+        # on a proper message object, so we do the same thing as the message create
+        # processor
+        message = self.cache.place_message_data(event.data)
```
LordOfPolls: This will result in the message deserializing twice. This effectively doubles the resource use of a bot that has prefixed commands

LordOfPolls: Perf branch mitigates this *slightly* due to its lazy deserializer, but its rather bad practice to be doing the same work twice. I'd be curious to see if its reliable to assume the message is already in the cache, as processors will always be triggered before listeners

AstreaTSS: I'll take a look - admittedly this uncertainty was the reason why I didn't rely on the cache in the first place, but there may be enough of a delay to make it work.

If it really comes down to it, the idea of making a really dumb loop to halt it while it waits for the message to appear is an option.

LordOfPolls: Instead of a polling loop, why not:
- check if the message is already in the cache
- if not, wait for a message create event
- continue

AstreaTSS: And done!

naff/client/client.py
```diff
@@ -1752,90 +1752,128 @@ async def _dispatch_interaction(self, event: RawGatewayEvent) -> None:
...
-                if message.content.startswith(prefix):
-                    prefix_used = prefix
-                    break
+        # if we didn't get a message, then we know we should wait for the message create event
+        if not message:
+            try:
+                # i think 10 seconds is a very generous timeout limit
+                event: MessageCreate = await self.wait_for(
+                    MessageCreate, checks=lambda e: int(e.message.id) == int(data["id"]), timeout=10
+                )
```
LordOfPolls: 10 seconds is a bit overkill. 2 seconds should always be the response target for an application that wants to feel responsive. Better to fail outright than to sit and wait for 10 seconds 
```suggestion
                # i think 2 seconds is a very generous timeout limit
                event: MessageCreate = await self.wait_for(
                    MessageCreate, checks=lambda e: int(e.message.id) == int(data["id"]), timeout=2
                )
```

AstreaTSS: Fair, I just picked a random number anyways.

